PR #3, opened 2021-01-28
HP-5 Set up CI

vkurup:
I started out with CircleCI, but am looking into Github Actions since CircleCI's integration instructions have left me stuck: https://github.com/CircleCI-Public/slack-orb/issues/230


.github/workflows/deploy.yml
```diff
@@ -0,0 +1,41 @@
...
+        run: |
+          python -m pip install --upgrade pip wheel pip-tools
+          pip-sync requirements/base/base.txt requirements/dev/dev.txt
+      - name: Login to Docker
+        run: |
+          inv aws.docker-login
+      - name: Build tag, push, and deploy image
+        run: |
+          [ "$GITHUB_REF" = refs/heads/main ] &&
+            ENVIRONMENT="production" ||
```
dchukhin, line 38: Does this mean that merging into the `main` branch would deploy the `main` branch to staging and production?

.github/workflows/test.yml
```diff
@@ -0,0 +1,54 @@
+name: Lint and Test code
+
+on:
+  pull_request:
+  schedule:
+    # run once a week on early monday mornings
```
dchukhin, line 6: Is this in 02:22 UTC?

dchukhin (approved):
It looks like it works; good job with it!
I left a couple of questions, but don't have any concerns about it

.github/workflows/deploy.yml
dchukhin: Question has been answered ✔️

vkurup: Hopefully (if my googling is correct), this is a ternary statement that will set ENVIRONMENT to "production" if we're on the main branch, else to "staging", which will then be inserted into the `inv` command below to deploy the proper environment

.github/workflows/test.yml
vkurup: Great question! we'll find out on Sunday :)

PR #5, opened 2021-02-01
HP-16 Clean up static files deployment settings

vkurup:
I think just adding the `favicon.ico` file was all we needed, but in the process of debugging, I cleaned up the static files handling (to the best of my understanding):

1. Add some things to .dockerignore to make the deployed image slightly smaller
2. Don't clear out STATIC_ROOT when calling collectstatic as it seems to clear out the files that webpack just put there
3. Remove some undefined tags from search_result.html which were causing warnings
4. Use whitenoise as recommended in both dev and deploy settings
5. Add a favicon.ico file from hip.phila.gov


Audiosutras (approved):
👍  Looks good vinod

PR #6, opened 2021-02-02
Add Slack notifications for deploys

vkurup:
We get automatic slack notifications for tests because Github includes it in the message for each new PR. But it doesn't for deploys, so this PR uses [an integration](https://github.com/marketplace/actions/slack-github-actions-slack-integration) to notify at the end of the deploy with the status.


dchukhin (approved):
🔔

PR #7, opened 2021-02-02
Don't include env vars in CI output

vkurup:
This takes advantages of new releases of django-k8s and invoke-kubesae which allow us to do deploys without leaking environment variables into the log output.


dchukhin (approved):
🚢

PR #11, opened 2021-02-08
[HP-10] Wagtail Static Page template (Report A Disease Page)

Audiosutras:
This PR adds wagtail models, templates, and styles that can be used to create the `Report A Disease` page seen in this [figma wireframe](https://www.figma.com/proto/OZAfVKxh8fs8XE7RLnlbY7/PDPH-COVID19?node-id=753%3A2&scaling=scale-down).  

* Make sure to make the report a disease `page` a child of the existing wagtail site page.
* Easiest way to re-create the `report-a-disease` page, is by navigating to the wireframe and copy/paste the content found on the wireframe's artboard into the cms.


apps/hip/templates/includes/right_scroll_nav.html
```diff
@@ -0,0 +1,25 @@
+{% load wagtailcore_tags %}
+<!-- Right Scroll Navigation for Anchor Links on desktop only
```
vkurup, line 2: Rather than using HTML comment syntax, we could use [django template multi-line comment syntax](https://docs.djangoproject.com/en/3.1/ref/templates/builtins/#comment) so that this doesn't get rendered.

apps/hip/templates/includes/right_scroll_nav.html
```diff
@@ -0,0 +1,25 @@
...
+
+  Example use:
+  <div class="columns">
+    <div class="column">other content column container</div>
+    {# {% include 'right_scroll_nav.html' %} #}
+  </div>
+
+  The column for the 'other content' div will
+  default to being is-two-quarter
```
vkurup, line 11: Wouldn't it be `is-three-quarter` since this one is `is-one-quarter`?

hip/static/styles/pages/base.scss
```diff
@@ -1,5 +1,13 @@
 @import "../colors.scss";
```
vkurup, line 1: Seems like sometimes we import without the `.scss` suffix and sometimes we import with it. I'm assuming that both work, but maybe we should stick with one? I think i'd slightly prefer including the suffix, to be more explicit, but am fine with either choice.

(Comment applies to multiple files, not just this location)

hip/static/styles/wagtail_styles.css
```diff
@@ -0,0 +1,4 @@
+.two-column-table__row > .field {
```
vkurup, line 1: Why is this one a `.css` file instead of `.scss`?

apps/hip/migrations/0003_auto_20210206_2119.py
```diff
@@ -0,0 +1,21 @@
+# Generated by Django 3.1.5 on 2021-02-06 21:19
```
vkurup, line 1: Can we squash these migrations into 1? (To do this, delete all the migration files and re-run `manage.py makemigrations`)

apps/hip/models.py
```diff
@@ -1 +1,76 @@
-# Create your models here.
+from wagtail.admin.edit_handlers import FieldPanel, StreamFieldPanel
+from wagtail.core import blocks
+from wagtail.core.fields import StreamField
+from wagtail.core.models import Page
+
+
+class TableRow(blocks.StructBlock):
+    column_1 = blocks.RichTextBlock(
+        max_length=255, required=False, help_text=("Text for column 1"),
```
vkurup, line 9: Looks like [black wants us to re-format this](https://github.com/caktus/philly-hip/runs/1855971661?check_suite_focus=true). Should be able to fix this by running `pre-commit run -a` locally.

vkurup (approved):
I haven't been able to fully get the report-a-disease page set up locally, but this looks like great so far. I made a few suggestions.

apps/hip/wagtail_hooks.py
```diff
@@ -0,0 +1,12 @@
...
+from django.utils.html import format_html
+
+from wagtail.core import hooks
+
+
+@hooks.register("insert_editor_css")
+def editor_css():
+    """Register the extra CSS styles to be used for the Wagtail editor."""
+    return format_html(
+        '<link rel="stylesheet" href="{}">', static("styles/wagtail_styles.css")
```
vkurup, line 11: Wonder if we could use `sass_processor()` and then wagtail_styles could be a scss file instead of css:
https://github.com/jrief/django-sass-processor#in-media-classes-or-properties

apps/hip/templates/includes/header.html
```diff
@@ -29,7 +29,7 @@
     <div class="navbar-end">
       <div class="navbar-item">
         <div class="buttons">
-          <a class="button is-size-7-touch is-link" href="#">
+          <a class="button is-size-7-touch is-link" href="/report-a-disease">
```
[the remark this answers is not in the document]
vkurup: I changed my mind about this. I think it's better to name the URLs the way we want them and then use the Wagtail Redirects functionality to point old URLs to new ones.

PR #12, opened 2021-02-08
Add wagtail user bar

vkurup:
... so logged in users can easily get to CMS admin

https://docs.wagtail.io/en/stable/topics/writing_templates.html#wagtail-user-bar


Audiosutras (approved):
👍 looks sharp!

PR #14, opened 2021-02-10
HP-33-pt-2 Move Back button inside column

vkurup:
Move the button inside the main column so that its click target doesn't get wiped out.


Audiosutras (approved):
👍

PR #16, opened 2021-02-12
HP-33-try-3 get back button working

vkurup:
https://caktus.atlassian.net/browse/HP-33

Using `javascript:history.back()` doesn't work if the user has clicked on right-nav links within the page. This instead uses the `HTTP_REFERER` header to get the actual last page that the user was on, defaulting to the home page URL if it is not present (like if the user came from an email link or something)


Audiosutras (approved):
👍  very nifty.

PR #17, opened 2021-02-13
Add citext installation to provisioning docs

vkurup:
... and remove things that have been either removed from jade-truffle recently, or are
not needed in our project.

The initial deploy of #15 failed because our DB user didn't have the permissions to install the `citext` extension on RDS. This adds provisioning docs to explain how we do that, so that we make sure to do it when we set up production.

I've also:
1. cleaned up some files and tasks that are not needed in our project (or have been [removed from jade truffle](https://github.com/caktus/jade-truffle/pull/20/files) recently)
2. Fixed the docs to sync media and DB from staging (allowing us to have local checkouts which match staging)


README.md
```diff
@@ -227,39 +227,33 @@ The basic command for resetting your local media is this:
...
-
-If you have already retrieved a database file, for example from a backup server, you can restore that dump using the
-``dump-file`` argument.
+To reset your local database from a deployed environment:
 
 
 ```sh
-    (hip)$ inv reset-local-db --dump-file="<PATH_TO_BACKUPFILE>"
+    (hip)$ inv staging pod.get-db-dump --db-var=DATABASE_URL
+    (hip)$ # make sure your local docker DB is running and configured in DATABASE_URL
```
Audiosutras, line 255: Could we add a `make` command for db restoration?

Audiosutras (commented):
Looks good Vinod! I didn't feel comfortable adding an approval. I think @dchukhin may better be able to comment on the invoke-kubsae changes.

tasks.py
```diff
@@ -23,101 +22,12 @@ def production(c):
...
-    Usage:
-        inv ci-run --command='pytest'
-    """
-    c.run(
-        f"docker-compose -f docker-compose.yml -f docker-compose-test.yml run --rm app sh -lc '{ command }'"
-    )
-
-
-@invoke.task
-def build_deployable(c):
```
dchukhin, line 49: I notice this and the `build_deploy` tasks are still in the `tasks.py` file in jade-truffle. Do we need them here too?

README.md
vkurup: Great thought! I'd prefer to move away from using the `Makefile` and use `inv` for everything (that's one of the purposes of the `invoke` tool: https://www.pyinvoke.org/#what-is-invoke). I'll see if I can add this as an invoke task.

tasks.py
vkurup: Can't sneak anything by you! I couldn't see a reason that we would need these on this project, but if you see value in them, I'd be happy to bring them back.

dchukhin: I don't _think_ we need them here. I guess we can make a mental note about the fact that we removed them.

dchukhin (approved):
💻

README.md
vkurup: Done!

PR #18, opened 2021-02-15
HP-32 Get right-scroll-nav to change colors appropriately

vkurup:
https://caktus.atlassian.net/browse/HP-32

... and add smooth scrolling for fun! :)
https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView

### DIFF
--- a/apps/hip/static/js/rightScrollNav.js
+++ b/apps/hip/static/js/rightScrollNav.js
@@ -1,17 +1,17 @@
 export default function () {
   /* Use IntersectionObserver to determine when a section of the page is within
-   * the user's view; when it is, add a CSS class to its respective nav title.
+   * the user's view; when it is, add a CSS class to its respective anchor tag.
+   * Docs: https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API
+   *       https://codepen.io/mishunov/pen/opeRdL
    */
 
-  const pageSections = document.querySelectorAll("section.page-section-hip");
-  const config = {
-    threshold: [0, 0.25, 0.5, 0.75, 1]
-  };
+  const pageSections = document.querySelectorAll("div.nav-heading-hip");
+  const config = {};
 
   // Define the IntersectionObserver.
   let observer = new IntersectionObserver(function (entries, self) {
     entries.forEach(entry => {
-      if (entry.isIntersecting && entry.intersectionRect.top > 0) {
+      if (entry.isIntersecting) {
         intersectionHandler(entry);
       }
     });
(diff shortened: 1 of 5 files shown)


Audiosutras (commented):
Looks great, however the section that becomes active in the right scroll navigation does so when that content section is first shown on the page. So if you are reading the `Why We Report` section and `HIPPA FAQS` becomes visible at the bottom of the page, then the navigation link shows `HIPPA FAQS` as being the active section. I think we need to wait until that section as at the top (or 15% from being at the top of the screen) before the navigation links changes.

Audiosutras (commented):
https://user-images.githubusercontent.com/37882933/108006430-71020700-6fc9-11eb-8485-ca8534df1246.mp4

This video is not perfect. `How to report` anchor link in the right navigation is not present. I think the end of the video best shows what I was thinking. When a user clicks an anchor link, the user is taken to the beginning of that section. I think when the beginning of a content section is towards the top of the page then the active link in the navigation should change to that section.

apps/hip/static/js/rightScrollNav.js
```diff
@@ -1,17 +1,17 @@
...
 
-  const pageSections = document.querySelectorAll("section.page-section-hip");
-  const config = {
-    threshold: [0, 0.25, 0.5, 0.75, 1]
-  };
+  const pageSections = document.querySelectorAll("div.nav-heading-hip");
+  const config = {};
 
   // Define the IntersectionObserver.
   let observer = new IntersectionObserver(function (entries, self) {
```
Audiosutras, line 12: I think adding the `options` argument while instantiating `IntersectionObserver` will do the trick. In particular, I think we need to play with `threshold`. https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API#creating_an_intersection_observer

apps/hip/static/js/rightScrollNav.js
```diff
@@ -23,22 +23,33 @@ export default function () {
...
-    // Give the 'is-active' CSS class to the nav title that is becoming active.
+    // Give the 'is-current-hip' CSS class to the nav title that is becoming active.
     if (elementBecomingActive) {
-      elementBecomingActive.classList.add("is-active");
+      elementBecomingActive.classList.add("is-current-hip");
     }
   }
+
+  // smooth scroll to #targets
+  document.querySelectorAll('a[href^="#"]').forEach(anchor => {
```
dchukhin, line 46: 🆒

dchukhin (commented):
Nice, I like the changes!
As Chris pointed out, some of the sections don't get highlighted at the correct time, so we'll need to adjust that. This is hard, though. I couldn't figure out the exact correct options/config to give the IntersectionObserver when I looked at it for a bit last week, though I believe that we can figure it out!

Audiosutras (approved):
👍 looks good.

PR #19, opened 2021-02-17
[HP-20] Add Home Page Part I

dchukhin:
**Note: this pull request depends on the HP-37 branch.**
If it's preferable, I can pull the relevant commits out though, so this can be reviewed separately

https://caktus.atlassian.net/browse/HP-20

In order to get this work to be reviewed earlier rather than later, this pull request:
 - adds a `HomePage` model with the following sections:
   - Quick Links
   - About
   - Recently Updated
 - adds a `get_most_recent_objects()` utility function for getting the most recently updated objects

The following still needs to be done (in a future pull request):
 - add a snippet for the 'Report A Disease' section
 - update the mobile styling for the 'Recently Updated' section


apps/hip/utils.py
```diff
@@ -0,0 +1,51 @@
...
+     - type_of_object
+     - model_name
+    Note: objects are returned in a list. It is also possible to combine querysets
+    together using queryset.union(), but it is difficult to do so for different
+    models without using .values(), which means that each object in the queryset
+    loses its methods. In order to preserve the methods, objects are combined in
+    a list.
+    """
+    # Get the most recent Page objects.
+    pages = Page.objects.exclude(latest_revision_created_at__isnull=True).annotate(
```
vkurup, line 28: I think we'll want `Page.objects.live()...` because it is currently retrieving draft pages too.

apps/hip/utils.py
```diff
@@ -0,0 +1,51 @@
...
+            When(file__endswith="pdf", then=Value("PDF", output_field=CharField())),
+        ),
+        model_name=Value("document", output_field=CharField()),
+    )[:object_count]
+
+    # Sort the combined_list by their 'updated_at' property.
+    combined_list = list(pages) + list(documents)
+    combined_list.sort(key=lambda x: x.updated_at, reverse=True)
+
+    # Return only the most recent object_count amount of objects.
```
vkurup, line 50: Very nice set of queries to get this info!

vkurup (commented):
This looks great! I had a couple questions/comments:
* I think I figured out how to get this page to show up at the root URL (by creating a new site, making a new HomePage, and making the new HomePage be the root page of the new site). Does that sound like what you were thinking?
* Even before doing all of that, the report-a-disease page content disappeared. Not a 404... I can still get to the page, but there's no content in there even though the page content is present in the DB. If I switch back to the develop branch, I can see it, but on this branch, it looks like this:
image

apps/hip/models.py
```diff
@@ -91,4 +91,71 @@ def get_context(self, request):
...
+        FieldPanel("about"),
+    ]
+
+    def get_context(self, request):
+        """Add recent_updates to context."""
+        from .utils import get_most_recent_objects
+
+        context = super().get_context(request)
+        context["recent_updates"] = get_most_recent_objects(object_count=10)
         return context
```
vkurup, line 161: This is a tricky thing to see in a github diff! this line used to be the last line of the StaticPage.get_context() method, but now it is only here and not there, so the StaticPage never returns a context. I'll push up a quick commit to fix

dchukhin: Good 👁️

apps/hip/utils.py
dchukhin: ✔️

apps/hip/utils.py
dchukhin: thanks!

vkurup (approved):

PR #20, opened 2021-02-17
HP-14 & HP-37 (PART 1) Header Desktop & Touch Devices

Audiosutras:
This PR updates the styles for the header on both desktop and touch devices. This PR also adds toggle functionality for the sidebar on touch devices. Also on touch devices, when a user clicks on the `search icon`, the search modal will be displayed for the user. By default the search modal shows what a populated search results would look like currently. In `mobile_search_modal.html`, set the filter and results sections to False to see what the default search modal would look like on touch devices. The search modal is not available for desktop, and at this time I'm not certain if there is a ticket for adding a search results page.

This PR does not address `report-disease` style updates. That will come in the part 2 pr for HP-37.

This PR also adds font awesome icons to the project.


apps/hip/static/js/sidebar.js
```diff
@@ -1,19 +1,38 @@
 export default function () {
-  const sideNavLinks = document.querySelectorAll(".sidenav-link-hip");
+  manageSideBarLinks();
+  closeButtonClicked();
```
vkurup, line 3: nitpick: I think I'd like this to be called `manageCloseButton` to match `manageSideBarLinks` (but you don't need to change it)

apps/hip/templates/hip/static_page.html
```diff
@@ -7,8 +7,7 @@
 
 {% block sidebar %}
   {# disable left sidebar but preserve the space that it occupies #}
-  <div class="column is-2 pr-0">
-  </div>
+  {% include 'includes/sidebar.html' with hidden_desktop=True %}
```
vkurup, line 10: Making this change removes the left sidebar completely (which also makes the comment above incorrect). If you're planning to fix that in the next PR, that is OK. Just want to make sure.

apps/hip/templates/includes/header.html
```diff
@@ -1,33 +1,47 @@
...
+    <div class="is-hidden-desktop is-full-width-hip is-full-height-hip columns push-to-back-hip is-mobile">
+      <div class="column is-2 is-jusify-content-flex-start">
+        <a role="button" id="navbar-burger" class="navbar-burger is-block is-size-5 mr-2 mt-4 is-mobile" aria-label="menu" aria-expanded="false" data-target="sidebarContent">
+          <span aria-hidden="true"></span>
+          <span aria-hidden="true"></span>
+          <span aria-hidden="true"></span>
+        </a>
+      </div>
+      <div class="column is-8">
+        <a class="logo-hip" href="https://www.phila.gov/">
```
vkurup, line 13: I know this is a post-sprint change, but let's make the logo go to `/home` since the client approved that change and it will make it easier for QA to move around the site by clicking the logo, rather than having to use the URL bar. (Can be done in a future PR)

vkurup (approved):
:framed_picture: 

This looks great! I made a few comments, but nothing that needs to block this from being merged, so Kat can start QAing.

apps/hip/templates/includes/header.html
Audiosutras: done

apps/hip/static/js/sidebar.js
Audiosutras: done

apps/hip/templates/hip/static_page.html
Audiosutras: I see, working on it.

Audiosutras: done

Audiosutras: Restored spacing on screens where the sidebar is hidden on desktop. User is still able to use the sidebar on touch devices.

PR #21, opened 2021-02-17
Lazy Load RightScrollNav.js 

Audiosutras:
This PR is to lazy load RightScrollNav.js. This script should only run on pages that use the right-scroll-nav html component; otherwise the Intersection Observer object raises an error.


hip/static/js/index.js
```diff
@@ -25,5 +25,20 @@ import RightScrollNav from "../../../apps/hip/static/js/rightScrollNav";
 document.addEventListener("DOMContentLoaded", function() {
   Header();
   SideBar();
-  RightScrollNav();
+  lazyLoad();
+
+  function lazyLoad() {
```
vkurup, line 30: Do we need this function? Can't we just do `lazyLoadRightScrollNav()` on line 28?

hip/static/js/index.js
```diff
@@ -25,5 +25,20 @@ import RightScrollNav from "../../../apps/hip/static/js/rightScrollNav";
...
+    return lazyLoadRightScrollNav()
+  };
+  function lazyLoadRightScrollNav() {
+    /* Lazy Load Right Scroll Nav Component Javascript
+    This javascript should only run on pages that use the right scroll nav component 
+    */
+    const rightScrollNavPages = ["report-a-disease"];
```
vkurup, line 37: Wonder if it would be possible to look for the `right-scroll-container-hip` div instead? That way we wouldn't have to keep this list of pages updated.

Audiosutras: I will try, I don't think `right-scroll-container-hip` is in the DOM on each page. I will need to find out the best method for checking that the element exists without throwing a reference error. Or may be we could use a try/catch block for this.

hip/static/js/index.js
Audiosutras: I don't think, I'm happy to remove it. I just did this in case there are other components that need to be lazy loaded in the future. I rather that they all in there on "lazy load" function than just in lazyLoad code block.

hip/static/js/index.js
vkurup: Exactly. That div is only on the pages where we want the JS to load, so it is the perfect thing to be our conditional check. Let me know if you want me to work on this, I'm sure I've done it before... just need to remember how :)

hip/static/js/index.js
vkurup: Hmm, even better would be for the lazy loading to happen within `RightScrollNav()` itself, rather than having `lazyload` in the name anywhere. The component is the best place to determine whether it should load since it has the knowledge of knowing what it needs to be able to function properly.

Audiosutras: done

hip/static/js/index.js
Audiosutras: done. Vastly simplified.

vkurup (approved):
I :heart: it!!

PR #22, opened 2021-02-18
Fix JS

vkurup:
2 things in this PR:
1. Bring back the docker-compose-deploy.yml file that allows you to do `inv image.up` to test an image locally
2. Delete the extra bundle that we were including in the template (and move the correct bundle down to the end of the DOM so it doesn't block page load)

I'm not 100% sure that this was the reason that JS was not working on staging, but I deployed this image and it looks like staging is working now


Audiosutras (approved):
👍

PR #25, opened 2021-02-22
[HP-43] Remove Home View

dchukhin:
https://caktus.atlassian.net/browse/HP-43

This pull request:
 - removes the `home` view and its template from the code base
 - updates the `StaticPage.get_context()` method to use a `HomePage`'s URL as a default URL for the `prev_url` context variable
 - adds a `get_home_page_url()` utility function to return the first live `HomePage`'s URL, or "/"
 - adds a `home_page_url` context processor to return a `home_page_url` in the context for each view
 - updates templates that referred to `/home` to use the `home_page_url` context variable
 - adds the `pytest-mock` library to developer requirements


apps/hip/views.py
```diff
@@ -12,7 +11,3 @@ def handler500(request, *args, **argv):
     response = render(request, "500.html", {})
     response.status_code = 500
     return response
-
-
-def home(request):
-    return TemplateResponse(request, "home.html")
```
vkurup, line 18: Should we remove the home.html placeholder template as well?

vkurup (approved):
one comment, then :house_with_garden:

PR #26, opened 2021-02-22
HP-12 Health Alerts page

vkurup:
https://caktus.atlassian.net/browse/HP-12

There are a lot of problems with this. Specifically, I did just the minimum to get styles working, instead of trying to learn the right way to do them. I DO want to learn the right way, but wanted to get this up for review sooner rather than later. It also needs tests, but I can work on that in the background while Kat is QAing, if we think this is close enough to get to QA. It will only really look/work well if we have a lot of health alerts in there, so I can help with that process on staging.

Here's a screenshot of what it looks like:
image

Other things that I didn't do yet:
* Filter widget
* Sign up page
* hover action which shows priority legend


apps/health_alerts/templates/health_alerts/health_alert_index_page.html
```diff
@@ -0,0 +1,96 @@
+{% extends "base.html" %}
+{% load wagtailcore_tags %}
+
+{% block title_suffix %}
+  {{ page.title }}
+{% endblock %}
```
Audiosutras, line 6: Naming the title of the health alerts index page `Health Alerts` seems to make the title of the page `Health Alerts Health Alerts`

Audiosutras (commented):
When the Health Alerts Index page is shown at `/health-alerts/`, when the page refreshes to show this page we need to update the is active section in the sidebar to be `Health Alerts`.

apps/health_alerts/models.py
```diff
@@ -0,0 +1,89 @@
...
+
+    class Priority(models.IntegerChoices):
+        UPDATE = 1
+        NOTIFICATION = 2
+        ADVISORY = 3
+        ALERT = 4
+
+    priority = models.IntegerField(choices=Priority.choices)
+
+    alert_date = models.DateField(default=datetime.date.today)
```
Audiosutras, line 51: This field currently is being displayed as `yyyy/mm/dd`. To be consistent with `Contact Us` snippet and other areas we are displaying a date in the admin. Should we use `mm/dd/yyyy`?

Audiosutras (approved):
💯 , in line with what you said in the PR. I think the client will be excited to see this page. I think there are a couple small style bugs that Kat will find and we can iterate on. For example, changing the icon for the filter dropdown. Links for the health alerts table and the right scroll nave should use `$dark-blue` etc.

I think we should merge the PR and then pick up the styles bugs as they come. Looks great.

PR #29, opened 2021-02-23
[HP-44] Limit Quick Links Cards To 4 Per Row

dchukhin:
https://caktus.atlassian.net/browse/HP-44

This pull request limits the number of quick links cards on the home page to 4 per row by beginning a new row every 4 quick links.

### DIFF
--- a/apps/hip/templates/hip/home_page.html
+++ b/apps/hip/templates/hip/home_page.html
@@ -7,18 +7,29 @@
 
 {% block content %}
   <div class="homepage-padding-hip">
-    <section class="columns">
-      {% for quick_link in page.quick_links %}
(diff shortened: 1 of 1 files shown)


vkurup (approved):
This works, but I think there might be a simpler method using just [Bulma's multiline directive](https://bulma.io/documentation/columns/options/#multiline). I haven't tested it here though, but could you try:
1. adding `is-multiline` to the element with the `columns` directive, and
2. adding `is-one-quarter` to the element with the `column` directive

I think that will make it so that each card is 1/4 of the width and will automatically go to the next line after 4.

If that doesn't work, then I approve this PR as-is.

dchukhin:
Closing, as we have resolved this in #32

PR #31, opened 2021-02-23
HP-60 - Remove modal-is-open-hip class from body element when sidebar modal is closed

Audiosutras:
https://caktus.atlassian.net/browse/HP-60


apps/hip/static/js/common.js
```diff
@@ -0,0 +1,9 @@
+export function mobileModalOpen (target) {
```
vkurup, line 1: Minor comments, but you don't have to accept them. I think I'd prefer these to be named `openMobileModal` and `closeMobileModal` which makes it clear to me that by calling this function I am going to perform that action. A comment here saying something like `// target: element to open as a modal` or something might be helpful as well.

vkurup (approved):
Minor comments, feel free to ignore. I like how you split the code out to reuse it!

Audiosutras:
> Minor comments, feel free to ignore. I like how you split the code out to reuse it!

Yea, I really like how we setup the javascript for this project. Sort of our own vanilla javascript version of react or vue.

apps/hip/static/js/common.js
Audiosutras: done

PR #32, opened 2021-02-23
HP-44-vinod Wrap quicklink cards at 4 per row

vkurup:
https://caktus.atlassian.net/browse/HP-44

Replacement for #29 

This uses Bulma's helpers to make the cards wrap at 4 cards per row, and changes the background color to match what is in the wireframe now.


dchukhin (approved):
🗂️

PR #33, opened 2021-02-23
Make modal's min-height be 100 percent of the viewport height, instea…

Audiosutras:
https://caktus.atlassian.net/browse/HP-51

This issue seems to only be reproducible on an actual mobile device. This can be accomplished by navigating to the hip site on your mobile phone.

My thinking is that modal needs to have a minimum height of 100 percent of the viewport's height. Originally this value was set to 100% of its parent container's height. I'm wondering if using `100vh` is more accurate.

### DIFF
--- a/hip/static/styles/_mixins.scss
+++ b/hip/static/styles/_mixins.scss
@@ -71,7 +71,7 @@
   bottom: 0;
   right: 0;
   width: 100%;
-  min-height: 100%;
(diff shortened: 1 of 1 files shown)


vkurup (approved):
I connected to my local dev by setting runserver to use 0.0.0.0:8000 and then connecting to my internal IP from my phone and this fix looks to work!

PR #34, opened 2021-02-23
HP-61 Fix updated date for internal pages on quicklink cards

vkurup:
https://caktus.atlassian.net/browse/HP-61

My bad... `date` is a method not an attribute


Audiosutras (approved):
👍

PR #35, opened 2021-02-23
HP-62 

Audiosutras:
Limit character length of the titles for quick link cards to 80 characters.
https://caktus.atlassian.net/browse/HP-62


apps/hip/models.py
```diff
@@ -163,7 +163,7 @@ def updated_date(self):
 
 class QuickLinkCard(blocks.StructBlock):
     title = blocks.CharBlock(
-        max_length=255,
+        max_length=80,
         required=True,
         help_text=("The linked text that will be visible to the reader"),
```
vkurup, line 168: Let's update the `help_text` to add "(maximum of 80 characters)"

vkurup (approved):

apps/hip/models.py
Audiosutras: done

PR #36, opened 2021-02-23
Update fonts to match wireframes

vkurup:
This:
1. adds montserrat and open-sans using the [docs from Google](https://fonts.google.com/specimen/Open+Sans?preview.text_type=custom&sidebar.open=true&selection.family=Open+Sans:ital@1).
2. Uses montserrat for buttons and open-sans for all other text
3. Makes text of buttons uppercase

### DIFF
--- a/hip/static/styles/includes/header.scss
+++ b/hip/static/styles/includes/header.scss
@@ -45,6 +45,8 @@
 .header-btn-hip {
   @include light-blue-bg;
   border-radius: 0;
+  text-transform: uppercase;
+  font-family: 'Montserrat', sans-serif;
 }
 
 .header-icon-hip {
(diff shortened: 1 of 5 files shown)


hip/static/styles/includes/header.scss
```diff
@@ -45,6 +45,8 @@
 .header-btn-hip {
   @include light-blue-bg;
   border-radius: 0;
+  text-transform: uppercase;
+  font-family: 'Montserrat', sans-serif;
```
Audiosutras, line 49: Should we also select the `button` element to apply line 48 and line 49 across all buttons. Or is this restricted to just the header buttons?

Audiosutras (approved):
👍

hip/static/styles/includes/header.scss
vkurup: Good point. I'll make sure it applies to all buttons

PR #37, opened 2021-02-23
HP-65 Only activate sidebar if element exists

vkurup:
https://caktus.atlassian.net/browse/HP-65


Audiosutras (approved):
👍 , This fixes HP-63 also

PR #38, opened 2021-02-24
HP 48 & HP 50

Audiosutras:
https://caktus.atlassian.net/browse/HP-48
https://caktus.atlassian.net/browse/HP-50

### DIFF
--- a/hip/static/styles/icons.scss
+++ b/hip/static/styles/icons.scss
@@ -130,14 +130,14 @@ To select that icon, do:
 
 // ------------------ Health Alert Priority icon colors --------- //
   .update-hip {
-    color: #3a833c;
+    color: $update;
   }
   .notification-hip {
-    color: #9400c6;
+    color: $notification;
   }
   .advisory-hip {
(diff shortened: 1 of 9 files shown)


hip/static/styles/icons.scss
```diff
@@ -127,3 +127,17 @@ To select that icon, do:
     content: fa-content($fa-var-external-link-alt);
   }
 }
+
+// ------------------ Health Alert Priority icon colors --------- //
+  .update-hip {
+    color: #3a833c;
```
vkurup, line 133: If we think we might use these in other places, we could move these to colors.scss (something I should have done in my original PR). Or we could wait to see if they are actually used elsewhere. Up to you.

vkurup (approved):
:rainbow:

hip/static/styles/icons.scss
Audiosutras: I'll add the colors to `_colors`, but keep the classes here.

PR #40, opened 2021-02-24
[HP-47] Align Snippet With Other Column

dchukhin:
https://caktus.atlassian.net/browse/HP-47

This pull request removes the `is-vcentered` CSS class from the parent `<div>` of both the snippet column and the 'About HIP' column, so the two columns align with each other vertically.

### DIFF
--- a/apps/hip/templates/hip/home_page.html
+++ b/apps/hip/templates/hip/home_page.html
@@ -21,7 +21,7 @@
       {% endfor %}
     </section>
 
-    <section class="columns is-vcentered is-desktop">
+    <section class="columns is-desktop">
       <div class="column is-6-desktop">
         <h3 class="is-size-4 pl-4"><strong>Report A Disease</strong></h3>
         {% contact_info %}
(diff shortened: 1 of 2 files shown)


vkurup (approved):
This works to get the tops aligned. I have a feeling that Kat will say that she wants the bottoms of the boxes aligned, like in the mockup, so I took a peek and I think this additional change will do that:
```
modified   apps/hip/templates/hip/home_page.html
@@ -28,5 +28,5 @@
       </div>
       <div class="column">
-        <div class="bg-lt-grey p-4 richtext-area-hip">
+        <div class="bg-lt-grey p-4 richtext-area-hip is-full-height-hip">
           {{ page.about|richtext }}
         </div>
modified   apps/hip/templates/tags/contact_info.html
@@ -17,5 +17,5 @@
       </div>
     </div>
-    <div class="columns is-vcentered is-mobile py-2 px-4 contact-container-hip">
+    <div class="columns is-vcentered is-mobile pt-2 px-4 contact-container-hip">
       <div class="column is-3 is-mobile dark-blue-bg-hip has-text-centered mr-1 is-full-height-hip">
         <span class="is-block moon-icon-hip is-size-3 pt-4"></span>
```

dchukhin:
Thanks for the review. Sure, I'll make those changes

PR #41, opened 2021-02-24
HP-68 Stripe tables ourselves so we can reset colors (and add spacing)

vkurup:
https://caktus.atlassian.net/browse/HP-68

### DIFF
--- a/hip/static/styles/pages/health_alert_index_page.scss
+++ b/hip/static/styles/pages/health_alert_index_page.scss
@@ -14,7 +14,7 @@
   border-spacing: 2px 0;
   border-collapse: separate;
   th {
-    background-color: $table-header-grey;
+    background-color: $dark-grey;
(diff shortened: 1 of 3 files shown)


hip/static/styles/pages/health_alert_index_page.scss
```diff
@@ -10,9 +10,23 @@
 
 .alert-table-hip {
   width: 100%;
+  border: 0;
+  border-spacing: 2px 0;
+  border-collapse: separate;
   th {
-    background-color: #323232;
-    color: #ffffff;
+    background-color: $table-header-grey;
```
Audiosutras, line 17: I think we should use $dark-gray to match the sidebar is-active link background. I put this in the PR for HP-48-50.

Audiosutras (approved):
👍

PR #44, opened 2021-02-24
Fix recent updates

vkurup:
I think the existing algorithm truncates documents and pages individually before sorting, so sometimes `combined_list` will not actually receive the latest list of pages and documents.

### DIFF
--- a/apps/hip/utils.py
+++ b/apps/hip/utils.py
@@ -33,7 +33,7 @@ def get_most_recent_objects(object_count=10):
             updated_at=F("latest_revision_created_at"),
             type_of_object=Value("PAGE", output_field=CharField()),
             model_name=Value("page", output_field=CharField()),
-        )[:object_count]
+        )
     )
     # Get the most recent Document objects.
     documents = Document.objects.annotate(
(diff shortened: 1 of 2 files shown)


apps/hip/utils.py
```diff
@@ -33,7 +33,7 @@ def get_most_recent_objects(object_count=10):
             updated_at=F("latest_revision_created_at"),
             type_of_object=Value("PAGE", output_field=CharField()),
             model_name=Value("page", output_field=CharField()),
-        )[:object_count]
+        )
```
dchukhin, line 36: Ah, alas, this used to have `.order_by("-updated_at")` before the `[:object_count]`, but it must have been erased.
I was thinking the `.order_by` would be faster than sorting in Python, but given the low number of objects, maybe it doesn't matter

dchukhin (approved):

apps/hip/utils.py
vkurup: Oops, hope I didn't accidentally delete the `order_by`. I think you're right about your approach being better (DB sorting rather than Python sorting). I'll go ahead and merge, but we can add that change to another PR later, if you'd like.

dchukhin: Sounds good. I think I must have erased it at some point in time

PR #47, opened 2021-02-25
HP-67

Audiosutras:
https://caktus.atlassian.net/browse/HP-67


dchukhin (commented):
This seems to work great for me in Firefox, but not in Chrome

dchukhin (commented):
👏 It looks like the long words are broken up!
Looking at the ticket, though, it also asked to reduce the white space between columns on mobile. Maybe mobile should have border-spacing of something like 1rem?

dchukhin (approved):
🐗

PR #48, opened 2021-02-25
[HP-49] Return QuickLinkCard UpdatedDate In Project Timezone

dchukhin:
https://caktus.atlassian.net/browse/HP-49

When a `QuickLinkCard` has a `link_page`, the `updated_date` is the `link_page`'s `latest_revision_created_at`'s date in the project timezone.

### DIFF
--- a/apps/hip/models.py
+++ b/apps/hip/models.py
@@ -1,4 +1,5 @@
 from django.db import models
+from django.utils.timezone import localtime
 
 from phonenumber_field.modelfields import PhoneNumberField
 from wagtail.admin.edit_handlers import FieldPanel, StreamFieldPanel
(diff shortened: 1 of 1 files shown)


vkurup (approved):
ahh, nice catch! I wonder if I made this bug appear by adding `.date()` to the end of this line. If I hadn't done that, then this method would have return at TZ-aware UTC datetime, and then the template would have automatically converted that to an America/New_York datetime, and then our date filter would have formatted that as a date properly.

dchukhin:
Perhaps. I wondered if I should pass the datetime into the template, but I guess since `updated_on` is a date now, we'll have to make sure to return the `latest_revision_created_at`'s date.

PR #50, opened 2021-02-27
HP-64: Health alerts legend, filter widget and design tweaks

vkurup:
I just noticed that on my new laptop, I had not configured tabs to be converted to spaces (:cone-of-shame:).

First commit: 
* converts tabs to spaces in the files I have committed

Second commit:
* Adds the hoverable legend on desktop only (as an include, so as not to clutter the template, and so it might be reused for the home page recent updates section if needed later)
* Aligns the year headings properly, makes them bold and gray
* Makes the table column widths match the mockup better (last 2 columns smaller than the first)
* Makes the table header slightly taller and vertical aligns the content inside of them
* Makes the topic 18px font, slightly larger than the other 2 columns
* Makes priority uppercase

Third commit:
* Styles the filter (no shadow, aligned properly with other items on the page)
* Adds a working JS-only filter that properly filters rows (including the year header if no matching rows in that year are desired), restripes the rows, and filters the right side links if needed

### DIFF
--- a/apps/health_alerts/static/js/healthAlertPriorityFilter.js
+++ b/apps/health_alerts/static/js/healthAlertPriorityFilter.js
@@ -8,8 +8,8 @@ export default function() {
       const allRows = document.querySelectorAll("[data-priority]");
       // we want to show any row that has the selectedPriority
       let rowsToShow = document.querySelectorAll(`[data-priority~="${selectedPriority}"]`);
-      if (selectedPriority === '0') {
-        // user chose '0', we will show all rows again
+      if (selectedPriority === 'All') {
+        // show all rows again
         rowsToShow = allRows;
       }
       // filter the rows, restripe them, and then filter the right side links
--- a/apps/health_alerts/static/js/healthAlertPriorityFilter.js
+++ b/apps/health_alerts/static/js/healthAlertPriorityFilter.js
@@ -1,0 +1,64 @@
+export default function() {
+  const selectElement = document.querySelector('.select-priority');
+
+  if (selectElement) {
+    // add a handler for the change event for the priority select dropdown
+    selectElement.addEventListener('change', (event) => {
+      const selectedPriority = event.target.value;
+      const allRows = document.querySelectorAll("[data-priority]");
+      // we want to show any row that has the selectedPriority
+      let rowsToShow = document.querySelectorAll(`[data-priority~="${selectedPriority}"]`);
+      if (selectedPriority === '0') {
+        // user chose '0', we will show all rows again
+        rowsToShow = allRows;
+      }
+      // filter the rows, restripe them, and then filter the right side links
+      filterRows(rowsToShow, allRows);
+      restripeRows(rowsToShow);
+      filterRightSideLinks(rowsToShow);
+    })
+  }
+
+  function filterRows(rowsToShow, allRows) {
+    // hide all the rows
+    allRows.forEach(row => row.hidden = true);
+    // show just the rows the user wants to see
+    rowsToShow.forEach(row => row.hidden = false);
+  }
+
+  function restripeRows(rows) {
+    const greyClass = "row-bg-grey-hip";
+    const whiteClass = "row-bg-white-hip";
+    let currentClass = whiteClass;
+    let nextClass = greyClass;
+
+    rows.forEach(row => {
+      if (row.dataset.year !== undefined) {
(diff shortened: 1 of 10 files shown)


apps/health_alerts/static/js/healthAlertPriorityFilter.js
```diff
@@ -0,0 +1,64 @@
...
+  const selectElement = document.querySelector('.select-priority');
+
+  if (selectElement) {
+    // add a handler for the change event for the priority select dropdown
+    selectElement.addEventListener('change', (event) => {
+      const selectedPriority = event.target.value;
+      const allRows = document.querySelectorAll("[data-priority]");
+      // we want to show any row that has the selectedPriority
+      let rowsToShow = document.querySelectorAll(`[data-priority~="${selectedPriority}"]`);
+      if (selectedPriority === '0') {
```
dchukhin, line 11: I wonder if we could make the choice values more meaningful, like 'all', 'alert', ...

dchukhin (commented):
Magic ✨

vkurup:
Thank @dchukhin, this is updated.

apps/health_alerts/static/js/healthAlertPriorityFilter.js
```diff
@@ -0,0 +1,64 @@
+export default function() {
```
Audiosutras, line 1: Awesome!

hip/static/styles/pages/health_alert_index_page.scss
```diff
@@ -10,22 +10,45 @@
...
   th {
     background-color: $dark-grey;
     color: $white;
     border-width: 0;
+    height: 3.5rem;
+    vertical-align: middle;
   }
+
   td {
     border-width: 0;
```
Audiosutras, line 25: I think we need to add `overflow-wrap: break-word` here.

apps/health_alerts/static/js/healthAlertPriorityFilter.js
```diff
@@ -0,0 +1,64 @@
...
+    selectElement.addEventListener('change', (event) => {
+      const selectedPriority = event.target.value;
+      const allRows = document.querySelectorAll("[data-priority]");
+      // we want to show any row that has the selectedPriority
+      let rowsToShow = document.querySelectorAll(`[data-priority~="${selectedPriority}"]`);
+      if (selectedPriority === 'All') {
+        // show all rows again
+        rowsToShow = allRows;
+      }
+      // filter the rows, restripe them, and then filter the right side links
```
Audiosutras, line 15: If `rowsToShow` is falsy, can we show a hidden div that has the text `No Health Alerts Found` or something like that?

Audiosutras (commented):
This PR looks great, and the popup is very nice. I left a few comments as enhancements.

apps/health_alerts/static/js/healthAlertPriorityFilter.js
Audiosutras: Also if possible, can we also hide the right scroll nav when rowstoShow is empty. When rowsToShow is Falsy, the nav headings are not visible, but the left border of the rightScrollNav is visible.

vkurup:
Thanks @Audiosutras! I've updated this to show a "no alerts found" element if there are no rows to show. I tried a lot of things to get the right scroll container to lose its border, but it didn't work. The border still shows if I set `el.hidden = true;` If I set `display: none` in CSS, then the entire element disappears, including the whitespace, which is too jarring.

At this point, I'd prefer to merge this and then maybe have you help me fix that issue separately, if needed. Having no alerts is unlikely to happen currently, but could happen once we add tagging, so we could fix it then as well.

Let me know if that is OK

apps/health_alerts/static/js/healthAlertPriorityFilter.js
```diff
@@ -24,6 +25,12 @@ export default function() {
     allRows.forEach(row => row.hidden = true);
     // show just the rows the user wants to see
     rowsToShow.forEach(row => row.hidden = false);
+    if (rowsToShow.length === 0) {
+      // if all rows are hidden, then show our special "no alerts found" row
```
Audiosutras, line 29: I prefer toggling a `is-hidden` or `is-hidden-hip` class. What do you think?

Audiosutras (approved):
👍

PR #53, opened 2021-03-01
HP-18 Disease Control Index Page (Part 1)

Audiosutras:
This PR adds the disease control index page. Only one disease control index page can exists at a time. After a disease control index page has been created, add a couple of child pages (DiseaseControlPage) in order to populate the index page with content.

This PR doesn't have the right scroll nav section. That will be implemented in part 2. 
https://caktus.atlassian.net/browse/HP-18


hip/static/styles/pages/disease_control_index_page.scss
```diff
@@ -0,0 +1,11 @@
+@import "../mixins";
+@import "../colors";
+
+.disease-control-index-page-hip {
+  margin: 0 3rem;
+  @include headers;
+  .disease-control-header {
```
vkurup, line 7: I can't find this class in our codebase. Is it something we can remove now, or is it needed for a future change you were planning?

vkurup (approved):
Looks good! A couple minor things that I am seeing:
image

* The "Disease Control" header and "Topic-specific Guidance" subheaders have a different font than the mockup (OpenSans vs Montserrat)
* The subpage looks like it needs to be indented a bit under each subheader

Unrelated to this PR (Just curious what you think about these items and whether we should create separate issues for them or fix them while you're working on this):
* should we unindent the selected left-sidebar item so that it matches the mockup?
* If there isn't much content, the page footer shows up in the middle of the page. Should we make it always go to the bottom of the viewport? (Or maybe we don't need to worry about that since real content will never be as short as my example?)

hip/static/styles/pages/disease_control_index_page.scss
Audiosutras: > Looks good! A couple minor things that I am seeing:
> image
> 
> * The "Disease Control" header and "Topic-specific Guidance" subheaders have a different font than the mockup (OpenSans vs Montserrat)
> * The subpage looks like it needs to be indented a bit under each subheader
> 
> Unrelated to this PR (Just curious what you think about these items and whether we should create separate issues for them or fix them while you're working on this):
> 
> * should we unindent the selected left-sidebar item so that it matches the mockup?
> * If there isn't much content, the page footer shows up in the middle of the page. Should we make it always go to the bottom of the viewport? (Or maybe we don't need to worry about that since real content will never be as short as my example?)

Thanks for the review! I didn't think about font's when working on this PR. When you opened your PR last week for changing the fonts I thought that it is what we should go with going forward. Happy to make the change to font-family; however, the font-family for the disease conditions page will differ from the font-family shown on the Report A Disease page (I think the wireframe for Report A disease page (the styling) is different than some of the other screens).

Audiosutras: Should they be consistent? What font family and styling should they have if they are to be consistent?

Audiosutras: > I can't find this class in our codebase. Is it something we can remove now, or is it needed for a future change you were planning?

I'm think this class maybe in Part 2 of HP-18. I have cherry-picked some commits from that PR.

Audiosutras: If the class is not found. I'll remove this.

vkurup: OK... my fonts PR included both Montserrat (for buttons and some headers) and Open Sans (for text)

PR #54, opened 2021-03-01
HP-66 Align Snippet 

Audiosutras:
https://caktus.atlassian.net/browse/HP-66


vkurup (commented):
This looks much better. The "Business Hours" one still overlaps the right edge a tiny bit on my phone:
Screenshot_20210303-093216

I *think* this is because the homepage has more horizontal padding on the outside than the report-a-disease page does.

Decreasing that padding seems to fix it for me (but I'm not sure if we need to make that change mobile-only or not... It still seems to look fine for me on desktop with this change):
```
modified   hip/static/styles/pages/home_page.scss
@@ -1,7 +1,7 @@
 @import "../variables";
 
 .homepage-padding-hip {
-  padding: 0.75rem 2rem;
+  padding: 0.75rem;
 }
 
 .quicklink-padding-hip {

```

Audiosutras:
> This looks much better. The "Business Hours" one still overlaps the right edge a tiny bit on my phone:
> Screenshot_20210303-093216
> 
> I _think_ this is because the homepage has more horizontal padding on the outside than the report-a-disease page does.
> 
> Decreasing that padding seems to fix it for me (but I'm not sure if we need to make that change mobile-only or not... It still seems to look fine for me on desktop with this change):
> 
> ```
> modified   hip/static/styles/pages/home_page.scss
> @@ -1,7 +1,7 @@
>  @import "../variables";
>  
>  .homepage-padding-hip {
> -  padding: 0.75rem 2rem;
> +  padding: 0.75rem;
>  }
>  
>  .quicklink-padding-hip {
> ```

Thanks @vkurup, changing it now.

vkurup (approved):
:straight_ruler:

PR #55, opened 2021-03-01
HP-76 (Add Right Scroll Nav - HP-18 Part 2)

Audiosutras:
https://caktus.atlassian.net/browse/HP-76


vkurup (approved):
I feel like the right scroll container should be pushed a little more to the right (to allow wider content in the main section, more like the report-a-disease page). We don't have a mockup for a right scroll section in this page yet, so that is just my personal opinion. :shipit: either way.

PR #57, opened 2021-03-02
HP-24 Part 2 - Style Disease and Conditions page & Add Breadcrumb

Audiosutras:
This PR:
1) Styles the `Disease and Conditions` page
2) Adds the breadcrumb component. This component is also added to the `Disease and Conditions` page and the `DiseasePage`.

https://caktus.atlassian.net/browse/HP-24

### DIFF
--- a/apps/disease_control/templates/includes/disease_control_card.html
+++ b/apps/disease_control/templates/includes/disease_control_card.html
@@ -16,7 +16,7 @@
           {% endif %}
         </p>
         {% if page.specific.is_emergent and not page.specific.description %}
-          <a class="button rounded is-block pt-2 emergent-btn-hip" href="{{ page.get_url }}">Emergent</a>
+          <a class="button is-rounded is-small mt-2 emergent-btn-hip" href="{{ page.get_url }}">Emergent</a>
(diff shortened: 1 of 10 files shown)


apps/disease_control/templates/includes/disease_control_card.html
```diff
@@ -1,10 +1,48 @@
...
+      </div>
+    {% else %}
+      <div class="page-description-hip column">
+        {{ page.description|richtext }}
+      </div>
+    {% endif %}
   </div>
+  {% if page.diseasepage and page.specific.description %}
+    <div class="columns">
+      {% if page.specific.is_emergent %}
```
vkurup, line 38: This `if` conditional should go inside the next `div`, otherwise this div gets hidden on non-emergent pages, so the `Updated` string ends up in the left column:
image

vkurup (approved):
one comment, then :shipit:

apps/disease_control/templates/includes/disease_control_card.html
Audiosutras: Thanks @vkurup. Pushed the change.

PR #58, opened 2021-03-02
HP-77 Fix navbar header on mobile

vkurup:
https://caktus.atlassian.net/browse/HP-77

This works OK, but I'm not 100% sure it is the right way to do this. I tried `position: sticky` as well, but couldn't get that to work.

Let me know if there are better solutions.

### DIFF
--- a/apps/hip/static/js/common.js
+++ b/apps/hip/static/js/common.js
@@ -2,10 +2,16 @@ export function openMobileModal (target) {
   // target: element to open as a modal
   target.classList.remove("is-hidden-touch");
   document.body.classList.add('modal-is-open-hip')
+  // un-stick the navbar
+  const navbar = document.getElementById("mobile-navbar-hip");
+  navbar.classList.remove("is-sticky-hip");
 };
 
 export function closeMobileModal (target) {
-  // target: element to open as a modal
+  // target: element representing the modal to close
   target.classList.add("is-hidden-touch");
   document.body.classList.remove('modal-is-open-hip')
+  // re-stick the navbar
+  const navbar = document.getElementById("mobile-navbar-hip");
+  navbar.classList.add("is-sticky-hip");
 };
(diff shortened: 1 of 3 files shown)


Audiosutras:
https://github.com/caktus/philly-hip/blob/8f662b22605608d877b149e4416fc7059ec01e40/hip/static/styles/includes/header.scss#L5-L9

1) Let's add this media query to `navbar-hip`
```scss
@media screen and (max-width: $breakpoint-tablet) {
    border-bottom: none;
}
```

https://github.com/caktus/philly-hip/blob/8f662b22605608d877b149e4416fc7059ec01e40/apps/hip/static/js/common.js
2) We need to update `openMobileModal` and `closeMobileModal`.
- `openMobileModal` we should remove the `sticky` classes
- `closeMobileModal` we should add the sticky classes

Otherwise, I think the feature is great.

vkurup:
Thanks @Audiosutras, this is updated now.

apps/hip/static/js/common.js
```diff
@@ -2,10 +2,16 @@ export function openMobileModal (target) {
...
+  navbar.classList.remove("is-sticky-hip");
 };
 
 export function closeMobileModal (target) {
-  // target: element to open as a modal
+  // target: element representing the modal to close
   target.classList.add("is-hidden-touch");
   document.body.classList.remove('modal-is-open-hip')
+  // re-stick the navbar
+  const navbar = document.getElementById("mobile-navbar-hip");
```
Audiosutras, line 15: I think we should change `navbar` to `header`, since there are multiple `nav` sections in the html code (header, sidebar, breadcrumbs`. Also I think we can move `const navbar = document.getElementById("mobile-navbar-hip");
` outside of the functions we are exporting and just definite it once above both function (I'm not certain about that though).

Audiosutras: I think there are a couple of small things we can open bug tickets for in QA that's not directly tied to this ticket.

Audiosutras (approved):
👍

apps/hip/static/js/common.js
vkurup: I'll update the name. I was afraid to move `navbar` outside the function because it looked to me like that would get imported on line 22 of `index.js` before the "DOMContentLoaded" event listener fired, so the DOM might not be built and ready yet.

PR #61, opened 2021-03-04
HP-85 Use breadcrumb navigation to set "is-active" sidebar link

Audiosutras:
https://caktus.atlassian.net/browse/HP-85

### DIFF
--- a/apps/hip/static/js/sidebar.js
+++ b/apps/hip/static/js/sidebar.js
@@ -20,7 +20,9 @@ export default function () {
       /* Use Breadcrumb navigation to determine which SideBarLink is active
 
       This function's logic is based on the assumption that sections that
-      have a page depth of greater than 1 will use breadcrumb navigation.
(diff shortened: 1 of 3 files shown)


apps/hip/static/js/sidebar.js
```diff
@@ -16,6 +16,25 @@ export default function () {
       }
     }
 
+    function getParentSidebarLink (currentPath) {
+      /* Use Breadcrumb navigation to determine which SideBarLink is active
+
+      This function's logic is based on the assumption that sections that
+      have a page depth of greater than 1 will use breadcrumb navigation.
```
dchukhin, line 23: Another assumption that could be added is that the first `<a>` of the breadcrumb will have an href that matches one of the sidebar `data-href` attributes

dchukhin (approved):
🔗

apps/hip/static/js/sidebar.js
Audiosutras: thanks and done.

PR #64, opened 2021-03-08
Fix sync-media-to-dev commands

vkurup:
The current sync commands puts the media folders inside "PROJECT/media/media" but our
code looks for files in "PROJECT/media". The latest version of invoke-kubesae allows you
to specify a `bucket-path` parameter which fixes that for us.


dchukhin (approved):
🛠️

PR #67, opened 2021-03-09
HP-89 Align View Alerts link

vkurup:
https://caktus.atlassian.net/browse/HP-89


Audiosutras (commented):
I think we should `is-align-items-center` or `is-align-items-middle`. I think `level` is a flex container, so this should work. Also something I noticed is that we will need to open a ticket a bug ticket for the responsiveness going from desktop to tablet. This is what I'm seeming.

Screen Shot 2021-03-09 at 2 49 42 PM

I don't think we need to fix that here, just wanted to make another person aware of it. If `align-items/justify-content` does not work. Let's set `margin-top` inside of a breakpoint media query that targets desktop devices and up.

Audiosutras (approved):
👍

PR #72, opened 2021-03-10
[HP-22] Emergency Response Page

dchukhin:
https://caktus.atlassian.net/browse/HP-22

This pull request:
 - adds an `EmergencyResponsePage`, which allows CMS users to add information about emergency responses
 - adds empty children pages (`VolunteerPage` and `HeatIndexPage`) for the `EmergencyResponsePage`
 - updates the JavaScript for adding an external link icon to also add a domain name for some external links (anything that is a child of an HTML element with the `external-links-with-domains-hip` CSS class)


vkurup (approved):
Nice work! I like the external links improvement. I think we need to do something to get the sidebar link to work and to get it to highlight appropriately when on this page, then :shipit:

PR #73, opened 2021-03-10
The big rename

vkurup:
This PR implements most of the changes we discussed here: https://docs.google.com/document/d/11Izn2zEZcWDZB6uQnrQIuOBJbjU68h3ijTnSIVu-pnI/edit

The one model I couldn't rename was DiseaseControlPage, because it is inherited by DiseaseAndConditionListPage and changing it triggered this bug: https://code.djangoproject.com/ticket/23521

I'm still open to feedback about whether this is a good idea (We don't have to do this!), but figure if we want to do it, the earlier the better, and best to do it before a sprint gets going.


dchukhin (commented):
Changes look fine to me, and tests pass, so that's a good sign.

I know the document also mentioned a `SimplePage` model, but I don't see it here. Were you planning on adding it to a future pull request, or does it make sense to add here?

I also keep thinking that the `ReportDiseasePage` is still pretty generic (except for the left and right hand navigation), so maybe there's a better name for it, but maybe not, since we only use it in 1 place 🤷‍♂️ . I'm not asking you to change it, just thinking out loud.

vkurup:
I didn't put `SimplePage` in yet because I wanted to discuss how "simple" it should be. I was thinking we might have a better sense of what features a SimplePage might need once we had done a few more sections.

Let's go ahead and revert ReportDiseasePage to be StaticPage for now, then. Maybe StaticPage will end up being our "SimplePage" template, as I could see it having features that they might want in a generic page.

apps/hip/models.py
```diff
@@ -105,6 +105,10 @@ class StreamAndNavHeadingBlock(blocks.StructBlock):
 class StaticPage(Page):
     """A Page with only sections of static content."""
 
+    max_count = 1
```
dchukhin, line 108: If we're keeping this as the `StaticPage`, should we remove the `max_count` and `parent_page_types` as well?

dchukhin (commented):
Potentially adding `SimplePage` later sounds good to me. I keep going back and forth about what to recommend about the `StaticPage`, so let's just go with what we have.
1 comment from me, then 🚢

vkurup:
@dchukhin Thanks for catching those issues. I just downloaded the DB from staging, ran these migrations and clicked around the interface and found a few other things that our tests did not catch. I *think* I've caught everything now, but could you take one more look and see if you think we should merge this?

dchukhin (approved):
Code looks good, and I clicked around the site locally, and it seems to work, so I say 👍

PR #74, opened 2021-03-11
[HP-52] Limit Documents To PDF, PNG, JPG

dchukhin:
https://caktus.atlassian.net/browse/HP-52

This pull request limits `Document` uploads to files that have either a .pdf, .png, or a .jpg extension. As a note, I find it surprising that the error only shows up after the document claims to have uploaded (step 2), but it's Wagtail's functionality, so I decided to not change it, and we can see how it does in QA.

### DIFF
--- a/hip/settings/base.py
+++ b/hip/settings/base.py
@@ -256,6 +256,8 @@ WAGTAILSEARCH_BACKENDS = {
     },
 }
 
+WAGTAILDOCS_EXTENSIONS = ["doc", "png", "jpg"]
+
 
 # Base URL to use when referring to full URLs within the Wagtail admin backend -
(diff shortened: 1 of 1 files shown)


hip/settings/base.py
```diff
@@ -256,6 +256,8 @@
     },
 }
 
+WAGTAILDOCS_EXTENSIONS = ["doc", "png", "jpg"]
```
vkurup, line 259: Don't we need `pdf` in here, or is that done automatically somehow? (haven't tested the PR out yet) :)

dchukhin: 🤦  saw that "doc" should have been "pdf", but forgot to commit it. Good 👁️

vkurup (approved):
:memo:

PR #76, opened 2021-03-12
Allow DiseasePage.description To Be Editable In Wagtail Again

dchukhin:
by adding it to the `DiseasePage`'s `content_panels`.


vkurup (approved):
:shipit:

PR #77, opened 2021-03-12
HP-96 Fix "sign up" button when screen shrinks

vkurup:
This change forces the `level-left` element to switch to `block: display` at screen widths below approx 1400px, which makes the "see more alerts" link move to another row. We have to decrease the margin a bit at those widths as well for this to work properly.

I also fixed a bug that I introduced in the health alerts page. I thought `p-6-desktop` would apply that padding to desktop pages, but that apparently doesn't work in bulma, so I moved those changes to a custom class.


dchukhin (approved):
I didn't realize bulma had such a class. 🆒

PR #78, opened 2021-03-12
HP-100 Move the header below the first year

vkurup:
https://caktus.atlassian.net/browse/HP-100

This moves the table header to be right after the first year, and it moves our "no alerts found" row to be just under the table header row as well. I don't think that there is any situation where the "no alerts row" will actually show up now on the main health alerts page.


dchukhin (approved):
🆗

PR #80, opened 2021-03-16
[HP-26] Add Login Page

dchukhin:
https://caktus.atlassian.net/browse/HP-26

This pull request:
 - adds a login page at the `/accounts/login/` URL, using the Django authentication system. Logging in redirects the user to the homepage.
 - updates links to the partner login to go to the `/accounts/login/` URL
 - adds a `previous_url` context processor, so previous URL can be determined on any page without repeating code
 - adds templates for resetting a password, mirroring the default Django functionality
 - sets the `request.session` expiration based on whether the user checks the "remember me" checkbox


Audiosutras (commented):
I think all of this applies to wider screens:
- can we set the `block content` section to have a `min-height` of `100vh` so that there is not white space beneath the footer
- Login button is no longer centered on wider screens and has a left alignment, should still be inline with Reset Password
Screen Shot 2021-03-17 at 12 05 29 PM

Not necessarily a part of the pull request. Can we add logic to the sidebar that will toggle a sign out button when there is an authenticated user

Audiosutras (approved):
👍

PR #81, opened 2021-03-16
HP-72 Implement Search

vkurup:
https://caktus.atlassian.net/browse/HP-72

A few things to note:
* Figma link in the JIRA ticket is old. I implemented this one: https://www.figma.com/file/OZAfVKxh8fs8XE7RLnlbY7/PDPH-COVID19?node-id=1083%3A28985 (I'll update JIRA when this PR is merged)
* Open to feedback about what other tests should be added
* This implements full text search on the selected fields, but not fuzzy search (no partial-matching)
* There are some limitations to the postgres search provided in Wagtail, but none that are deal-breakers, I don't think: https://docs.wagtail.io/en/stable/reference/contrib/postgres_search.html#postgres-search
* Most of the frontend/JS changes are due to the fact that the search mockup is a modal, but only on mobile. So I had to make it so that if the user clicks around the various pagination links, they can still close the modal and be taken back to the page that they were initially on.


apps/hip/templates/includes/search_result_card.html
```diff
@@ -1,19 +1,30 @@
...
+            <div class="level-item search-page-priority-badge-hip {{ result.specific.get_priority_display|slugify }}-hip">
+              <i class="fa {{ result.specific.get_priority_icon }}"></i>
+              {{ result.specific.get_priority_display|upper }}
+            </div>
+          {% endif %}
+        </div>
       </div>
+
+      {% if result.search_description %}
+        <p>{{ result.search_description|safe }}</p>
```
dchukhin, line 26: Should this field have any values? After I ran the `update_index` command locally, this field is still blank, so nothing shows up here for any of the cards.

apps/hip/templates/includes/search_pagination.html
```diff
@@ -0,0 +1,41 @@
+{% if search_results.has_other_pages %}
```
dchukhin, line 1: Nice; this file is very readable

dchukhin (commented):
Great work on it!
I like how it works, and I'm sure we can make improvements in the future if they are requested. Fuzzy matching would be nice, but I don't know how much people would care about it (probably not enough for the amount of work to add it).
I don't think we need to add tests for each model's `search_fields`, unless there was a field we explicitly didn't want to search on, so what you have should be fine.

I did find 2 🐛s, which I'll mention here. If you'd rather fix them in a separate pull request, feel free to do that too.

:bug: # 1:
 - search for something that gets no results (like skdjfnsjkdfnsjdhf)
 - search for something else that gets no results (like kjsdfskjdfnjs)
 - click back
 - click back
 - click back
 - notice that clicking back keeps you going back and forth between the 2 results pages with no results

🐛 # 2:
 - go to the Report A Disease page
 - search for something (like ‘Philadelphia’)
 - click the x to close the modal
 - notice that you’re now on the homepage (not where you were prior to searching)

apps/hip/templates/includes/search_result_card.html
vkurup: Oops, I meant to mention this in the PR description, but forgot. The [`search_description`](https://docs.wagtail.io/en/stable/reference/pages/model_reference.html#wagtail.core.models.Page.search_description) field is available in the "Promote" tab for all pages. Although, I believe the documentation there is incorrect. It does NOT seem to be indexed.

My feeling is that we should recommend that editors always update the search_description field with the content that they want to see in the search result card. We might be able to do a fallback where if search_description is not present we look for another field (like 'description'?) but not all models will have that field. The question I have is whether we should INDEX the search_description field. It is not currently indexed, which makes sense to me since it is not shown on the page, and I don't like the idea of a page showing up in results for a term, but that term not being present on the page. But I suppose we could put it in the `<meta name="description">` field... thoughts?

apps/search/tests/test_views.py
```diff
@@ -0,0 +1,53 @@
...
+
+from apps.common.utils import get_home_page_url
+from apps.disease_control.tests.factories import DiseaseControlPageFactory
+
+
+def test_search_no_query(client, db):
+    url = reverse("search")
+    response = client.get(url)
+    assert response.status_code == 200
+    context = response.context
```
dchukhin, line 11: Does this `context` variable get used (here and in other tests)?

apps/common/models.py
```diff
@@ -9,3 +13,42 @@ class IndexedTimeStampedModel(models.Model):
...
+    ]
+
+    promote_panels = [
+        # This MultiFieldPanel is copy-pasted from wagtail.admin.edit_handlers so we can override
+        MultiFieldPanel(
+            [
+                FieldPanel("slug"),
+                FieldPanel("seo_title"),
+                FieldPanel("show_in_menus"),
+                # override Page's promote_panels to remove search_description (since we
```
dchukhin, line 45: 👍 on the comment

dchukhin (approved):
1 small comment, and 🚢

PR #82, opened 2021-03-16
HP-13 - Health Alerts Sign Up Form

Audiosutras:
This PR:
- Adds Health Alerts Sign Up Form, presented as a modal although the user is directed to the sign up page when the `sign up for alerts` button is clicked
- Using the django messages framework the user is shown a success message after the sign up instance is successfully saved to the database
- HealthAlertsSignUp has been added to the django admin with the `verbose_name` of `Health Alerts Subscribers`. 
- A few tests for the django view and form have been added.


apps/health_alerts/models.py
```diff
@@ -108,3 +112,47 @@ def get_priority_color(self):
 
     def serve(self, request):
         return redirect(self.alert_file.url)
+
+
+class HealthAlertsSignUp(models.Model):
```
vkurup, line 117: Nitpick. I'd prefer the name `HealthAlertSubscriber` since that matches what each instance is and I like using singlar terms in model names better than plurals. (I got the idea by seeing the names you chose for the `verbose_name` below and felt that matched better)

apps/health_alerts/constants.py
```diff
@@ -0,0 +1,24 @@
+ANIMAL_VETERINARY_CLINICS = "AV"
```
vkurup, line 1: Another way to do this is to use Django's enumeration types and include this a class inside the model class, which encapsulates things a little better IMO: https://docs.djangoproject.com/en/3.1/ref/models/fields/#enumeration-types

But it is fine to do it like this too.

apps/health_alerts/models.py
```diff
@@ -108,3 +112,47 @@ def get_priority_color(self):
...
+    """Health Alert Sign Up
+
+    This model is a standard django model used
```
vkurup, line 120: I would amend to mention something like "Those newsletters are managed completely outside our system."

apps/health_alerts/templates/health_alerts/health_alerts_sign_up.html
```diff
@@ -0,0 +1,89 @@
...
+      <h5 class="is-size-5 pt-2 pl-4">
+        <strong>Health Alerts Sign Up Form</strong>
+      </h5>
+      <button id="health-alert-signup-close-btn-hip" data-href="/health-alerts/" class="button btn-no-border-hip pt-4">
+        <i class="close-icon-hip"></i>
+      </button>
+    </div>
+    <div class="px-4 columns is-widescreen">
+    {# Sign Up Form Section #}
+    <form class="column is-half-desktop class="column" action="/health-alerts-subscriber-signup" method="post">
```
vkurup, line 18: I think there's an extra `class="column` in there ^

apps/health_alerts/tests/test_views.py
```diff
@@ -0,0 +1,68 @@
...
+    )
+
+
+@pytest.fixture
+def url():
+    return reverse("health_alerts_sign_up")
+
+
+def test_get_sign_up_page(db, settings, client, url):
+    settings.STATICFILES_STORAGE = (
```
vkurup, line 32: You should be able to add this change to your dev settings and then remove the settings fixture here:
https://github.com/caktus/philly-hip/pull/81/files?file-filters%5B%5D=.py#diff-f35ca6429ddf447226fc2ff1342c932f17948d02069c2f59819f9af4c28273fa

hip/settings/base.py
```diff
@@ -198,6 +198,8 @@
 AWS_ACCESS_KEY_ID = os.getenv("AWS_ACCESS_KEY_ID")
 AWS_SECRET_ACCESS_KEY = os.getenv("AWS_SECRET_ACCESS_KEY")
 
+# Messages
+MESSAGE_STORAGE = "django.contrib.messages.storage.session.SessionStorage"
```
vkurup, line 202: We shouldn't need this right? If we don't set it, the messages framework will use cookies first and then fallback to sessions if necessary: https://docs.djangoproject.com/en/3.1/ref/contrib/messages/#storage-backends

hip/static/js/index.js
```diff
@@ -24,6 +24,8 @@ import SideBar from "../../../apps/hip/static/js/sidebar";
 import RightScrollNav from "../../../apps/hip/static/js/rightScrollNav";
 import MobileSearchModal from "../../../apps/hip/static/js/mobileSearchModal";
 import HealthAlertFilter from "../../../apps/health_alerts/static/js/healthAlertFilter";
+import HealthAlertSignup from "../../../apps/health_alerts/static/js/healthAlertSignup";
+import healthAlertMessage from "../../../apps/health_alerts/static/js/healthAlertMessage";
```
vkurup, line 28: Curious if there is a reason this one starts with a lowercase, unlike the others?

Audiosutras: Nope, just was rushing to get the thought and didn't correct spelling to fit the convention. Will do.

Audiosutras: done

apps/health_alerts/models.py
Audiosutras: sounds good! making the change.

Audiosutras: done

apps/health_alerts/constants.py
Audiosutras: Thanks, will do this going forward!

Audiosutras: Implemented. done.

apps/health_alerts/templates/health_alerts/health_alerts_sign_up.html
Audiosutras: I think I didn't indent this section properly.

apps/health_alerts/tests/test_views.py
Audiosutras: done

hip/settings/base.py
Audiosutras: done

hip/settings/dev.py
```diff
@@ -3,6 +3,8 @@
 from .base import *  # noqa
 
 
+STATICFILES_STORAGE = "django.contrib.staticfiles.storage.StaticFilesStorage"
```
vkurup, line 6: This needs to be inside of a conditional so that it is applied only during tests (we want to use whitenoise during dev to make sure that we see as many prod issues as possible during local development). So something like this:
```
if "pytest" in sys.modules:
    STATICFILES_STORAGE = "django.contrib.staticfiles.storage.StaticFilesStorage"
```

vkurup (approved):
One minor change and then :shipit:

hip/settings/dev.py
Audiosutras: done

PR #83, opened 2021-03-16
Quick fix for CI

vkurup:


Audiosutras (approved):
🍎

webpack-stats.json
```diff
@@ -0,0 +1 @@
+{"status":"done","chunks":{"main":[{"name":"main.js","path":"/home/vkurup/dev/philly-hip/hip/static/bundles/main.js"}]},"error":"ModuleBuildError","message":"Module build failed (from ./node_modules/babel-loader/lib/index.js):\nError: ENOENT: no such file or directory, open '/home/vkurup/dev/philly-hip/apps/health_alerts/static/js/healthAlertPriorityFilter.js'"}
```
dchukhin, line 1: Does the path matter here (since I'm guessing `/home/vkurup/dev/philly-hip/hip/static/bundles/main.js` won't be accurate for CI or other environments)?

dchukhin: Actually, running this on my 💻 updates the contents of `webpack-stats.json`, which doesn't break anything, but seems like we'll end up having many meaningless updates to this file either accidentally committed or carefully not committed. Now that it's in the repo, maybe we can re-add this file to the `.gitignore` file?

vkurup: That's a great suggestion! Right, the content of this file is meaningless, but "something" is needed here for CI to pass. The longer term fix might be to run `npm install && npm run build-dev` on CI, but that seems like a lot of work just to get this single file up there

PR #84, opened 2021-03-17
HP-103 Break down functions more discretely

vkurup:
https://caktus.atlassian.net/browse/HP-103

...to prevent doing things on the Disease&Condition page that we only want to do on the
main Health Alerts page.


dchukhin (approved):
💻

PR #85, opened 2021-03-17
HP-98

Audiosutras:
https://caktus.atlassian.net/browse/HP-98


apps/disease_control/models.py
```diff
@@ -179,7 +180,9 @@ class EmergentHealthTopicListPage(Page):
 
     def get_context(self, request):
         context = super().get_context(request)
-        context["ordered_diseases"] = DiseaseAndConditionDetailPage.objects.filter(
-            is_emergent=True
-        ).order_by("title")
+        context["ordered_diseases"] = (
+            DiseaseAndConditionDetailPage.objects.filter(is_emergent=True)
+            .annotate(updated_at=F("latest_revision_created_at"))
```
dchukhin, line 185: Any reason to not just do `.order_by("-latest_revision_created_at")` here?

apps/disease_control/models.py
```diff
@@ -179,7 +180,9 @@ class EmergentHealthTopicListPage(Page):
 
     def get_context(self, request):
         context = super().get_context(request)
-        context["ordered_diseases"] = DiseaseAndConditionDetailPage.objects.filter(
-            is_emergent=True
-        ).order_by("title")
+        context["ordered_diseases"] = (
```
dchukhin, line 183: Could we also add a test that the objects in `context["ordered_diseases"]` are ordered by their `latest_revision_created_at` field?

apps/disease_control/models.py
Audiosutras: No, just trying to be consistent with what was done in `hip.utils` for the `recently updated` table. Happy to simplify if you think that would be better.

Audiosutras: https://github.com/caktus/philly-hip/blob/25f897e44ed80227eb315a619b92a2bc29732038/apps/hip/utils.py#L26-L37

dchukhin: Makes sense. I think the main reason for the `updated_at` annotation in the utils.py method was because we wanted to call the same field name (`updated_at`) in the template even though it came from different fields in different models (`Page.latest_revision_created_at` and `Document.created_at`). Since it saves a little computing time to not do the annotation, and since we can just use `Page.latest_revision_created_at` in the template here, I'd vote to not add the annotation here.

Audiosutras: Ah Got it. Changed.

apps/disease_control/models.py
Audiosutras: Pushed a test.

dchukhin (approved):
🉑

PR #88, opened 2021-03-18
[HP-109] Fix Health Alert Signup On Disease Page

dchukhin:
https://caktus.atlassian.net/browse/HP-109

This pull request:
 - updates the link from the disease detail page to the health alerts signup page to be the real link
 - updates other template files that have a link to the health alerts signup page to use Django's `url` template tag
 - adds a trailing slash at the end of the `health_alert_subscriber` URL path


vkurup (approved):
Good catch on the trailing slash!

PR #89, opened 2021-03-18
[HP-112] Disaply Link Icon For Non-Health-Alert Pages

dchukhin:
https://caktus.atlassian.net/browse/HP-112

This pull request adds the link icon and the word 'PAGE' to each card in the search results that is not a health alert page.

### DIFF
--- a/apps/hip/templates/includes/search_result_card.html
+++ b/apps/hip/templates/includes/search_result_card.html
@@ -14,7 +14,7 @@
         </div>
         <div class="level-right">
           {% if result.specific.get_priority_display %}
-            {# Health alert pages show the icon for their priority and their priority #}
+            {# Health alert pages show the icon for their priority and the word to describe their priority #}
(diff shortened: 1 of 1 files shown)


apps/hip/templates/includes/search_result_card.html
```diff
@@ -14,10 +14,17 @@
         </div>
         <div class="level-right">
           {% if result.specific.get_priority_display %}
+            {# Health alert pages show the icon for their priority and their priority #}
```
Audiosutras, line 17:
```suggestion
            {# Health alert pages show the icon for their priority  #}
```

Audiosutras: You may have intended `and their priority`; I wasn't certain though.

Audiosutras (approved):
👍

apps/hip/templates/includes/search_result_card.html
dchukhin: I did actually mean to include both (their priority and the icon for their priority). I'll try to reword it.

PR #90, opened 2021-03-18
[HP-113] Add Padding Between Icon And Text In Search Results Cards

dchukhin:
https://caktus.atlassian.net/browse/HP-113


Audiosutras (approved):
👍

PR #91, opened 2021-03-18
add short_description to HomePage model and update template

Audiosutras:
https://caktus.atlassian.net/browse/HP-87


apps/hip/templates/hip/home_page.html
```diff
@@ -7,6 +7,14 @@
 
 {% block content %}
   <div class="homepage-padding-hip">
+    {% if page.title %}
+      <section class="pb-4 px-2">
+        <h3 class="is-size-3"><strong>{{ page.title }}</strong></h3>
```
Audiosutras, line 12:
```suggestion
        <h2 class="is-size-2"><strong>{{ page.title }}</strong></h2>
```
I think this might be the more appropriate size. Let me know if you agree and we accept the suggestion.

dchukhin: Since it's the page title, it seems like it should have the highest priority of headings. Since it looks like we use h2 for titles on other pages, I support this being an h2.

dchukhin (approved):
👍 to the h2 suggestion, and looks good

PR #92, opened 2021-03-18
Use Django's Url Template Tag For Login & Logout URLs In Templates

dchukhin:
This pull request changes hard-coded login & logout URLs in templates to use Django's `url` template tag with the URL's name.


Audiosutras (approved):
👍

PR #94, opened 2021-03-18
HP-110

Audiosutras:
https://caktus.atlassian.net/browse/HP-110

### DIFF
--- a/apps/health_alerts/forms.py
+++ b/apps/health_alerts/forms.py
@@ -4,6 +4,8 @@ from .models import HealthAlertSubscriber
 
 
 class HealthAlertSubscriberForm(ModelForm):
+    use_required_attribute = False
+
     class Meta:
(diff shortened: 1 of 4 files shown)


apps/health_alerts/forms.py
```diff
@@ -4,6 +4,8 @@
 
 
 class HealthAlertSubscriberForm(ModelForm):
+    use_required_attribute = False
```
dchukhin, line 7: I didn't know about this attribute
🆒

dchukhin (approved):
📄

apps/health_alerts/forms.py
Audiosutras: yea, @vkurup and I were discussing `IntegerChoices` and `TextChoices` fields that were added in Django 3.1 that made me look for something like the above for forms in the new django version. I fortunately found the above.

PR #95, opened 2021-03-19
[HP-111] Make Search Icon Clickable

dchukhin:
https://caktus.atlassian.net/browse/HP-111

This pull request makes the search icon clickable both in the header and on the search modal.
Based on the team discussion about removing the search bar, I left the search icon alone on the search results page.

### DIFF
--- a/apps/hip/templates/includes/header.html
+++ b/apps/hip/templates/includes/header.html
@@ -30,15 +30,17 @@
   <div class="column is-four-fifths columns is-desktop mobile-hidden-hip">
     <div class="column is-desktop is-hidden-touch is-6 pt-5">
       <div class="navbar-item is-full-height-hip">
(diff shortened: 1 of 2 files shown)


Audiosutras (commented):
Screen Shot 2021-03-19 at 9 52 15 AM

The search icon for the search input that is a part of the search results page does not also work.

Audiosutras:
> Screen Shot 2021-03-19 at 9 52 15 AM
> 
> The search icon for the search input that is a part of the search results page does not also work.

Based on the consensus to your question @dchukhin  from standup, we want to remove the search input that's located on the search results page.

dchukhin:
Sounds good. I believe that will be done in HP-118

Audiosutras (approved):
👍

Audiosutras:
> Sounds good. I believe that will be done in HP-118

Yep! Good to go.

PR #96, opened 2021-03-19
HP-115

Audiosutras:
https://caktus.atlassian.net/browse/HP-115

### DIFF
--- a/apps/disease_control/templates/disease_control/disease_and_condition_list_page.html
+++ b/apps/disease_control/templates/disease_control/disease_and_condition_list_page.html
@@ -2,7 +2,9 @@
 {% load wagtailcore_tags %}
 
 {% block content %}
-{% include 'includes/breadcrumb.html' with is_hidden=breadcrumb_is_hidden %}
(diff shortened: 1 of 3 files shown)


apps/disease_control/templates/disease_control/disease_and_condition_list_page.html
```diff
@@ -2,10 +2,10 @@
 {% load wagtailcore_tags %}
 
 {% block content %}
-{% include 'includes/breadcrumb.html' %}
+{% include 'includes/breadcrumb.html' with is_hidden=breadcrumb_is_hidden %}
```
vkurup, line 5: How about just not calling this `include` at all? So something like:
```
{% if not breadcrumb_is_hidden %}
  {% include 'includes/breadcrumb.html' %}
{% endif %}
```

Audiosutras: Done. Thanks

vkurup (approved):
:baguette_bread:

PR #97, opened 2021-03-19
Make Logout Button Icon Show Up In Template

dchukhin:
This pull request makes the logout icon show up in the template by using the correct mixin in the `icons.scss` file. Also, the icon now has a little padding, in order to match the wireframe.


Audiosutras (approved):
👍

PR #99, opened 2021-03-19
Hp 119

Audiosutras:
https://caktus.atlassian.net/browse/HP-119


apps/health_alerts/tests/test_forms.py
```diff
@@ -20,6 +20,15 @@ def subscribe_data():
     )
 
 
+def test_asterisks_marks_required_fields(db):
```
Audiosutras, line 23:
```suggestion
def test_asterisks_mark_required_fields(db):
```

dchukhin (approved):
This does work and fix the issue *️⃣ .
I wonder if it would be easier to add something like `{% if field.field.required %}*{% endif %}` to the label in the template instead. I'm not sure we need to change it; I'll leave that up to you.

PR #100, opened 2021-03-22
[HP-118] Remove Second Searchbar

dchukhin:
https://caktus.atlassian.net/browse/HP-118

This pull request:
 - removes search bar on search results page
 - adds the value of the search query into the search bar input in the header


Audiosutras (approved):
🍎

PR #101, opened 2021-03-23
HP-120

Audiosutras:
https://caktus.atlassian.net/browse/HP-120

Sidebar being hidden for the login screen is slightly different than it being hidden for the report a disease page. Dev's are now able to pass `hidden_column_desktop=True` to hide the column and right border of the sidebar on desktop devices.


dchukhin (approved):
🤾

PR #103, opened 2021-03-23
HP-99 Show user error message before file gets added to repository

vkurup:
https://caktus.atlassian.net/browse/HP-99


Audiosutras (approved):
👍 , looks good.

PR #105, opened 2021-03-24
[HP-90] PCW MSA Homepage

dchukhin:
https://caktus.atlassian.net/browse/HP-54

This pull request:
 - adds a `PCWMSAHomePage` page
 - adds a method to `HipBasePage` for determining if a page is a PCWMSA page or not
 - adds a `get_pcwmsa_home_page_url()` utility function to return the `PCWMSAHomePage` URL
 - redirects some users to the PCW MSA homepage when they log in, according to the following order of precedence: if they are in the 'Closed POD' `Group`, then they go to the `ClosedPODHomePage`. Else, if they are in the PCW MSA `Group`, then they go to the `PCWMSAHomePage`. This means that if a user is in both `Group`s, the Closed POD `Group` takes precedence)
 - updates the left sidebar to show different options when the user is on a PCW MSA page

Note for testing: users are only redirected to the PCW MSA home page upon login if they are in the 'PCW MSA' Django `Group`. To test this functionality, add yourself to the 'PCW MSA' `Group` (through the wagtail interface or the command line).


apps/auth_content/models.py
```diff
@@ -109,3 +109,34 @@ def get_context(self, request):
         context["closedpod_children_pages"] = self.get_siblings()
         context["closedpod_home_url"] = self.get_parent().url
         return context
+
+
+class PCWMSAHomePage(StaticPage):
+    max_count = 1
```
vkurup, line 115: Wonder if we should limit parent types for this (and the closedpod page) to the Home Page?

vkurup (approved):
:capital_abcd:

apps/auth_content/models.py
dchukhin: Sure, sounds like a good idea

PR #106, opened 2021-03-24
HP-122 & HP-124

Audiosutras:
- Remove back button and add sidebar back to desktop

As a team we have decided to keep the `sidebar` as visible on the `login.html` page and the `password_reset.html` page. The `< Back` button has been removed from the `login.html`. Both the `login.html` and `password_reset.html` page should fully resemble each other.


dchukhin (approved):
🆗

PR #107, opened 2021-03-24
HP-123

Audiosutras:
- remove sticky-header when health alerts signup form is open.
- https://caktus.atlassian.net/browse/HP-123


dchukhin (approved):
🆗

PR #108, opened 2021-03-26
[HP-128] Show Authenticated Home Page Links In Sidebar

dchukhin:
https://caktus.atlassian.net/browse/HP-128

This pull request adds links to the sidebar for authenticated home pages (`ClosedPODHomePage` and `PCWMSAHomePage`). Each of these links should only show up when:
 - the user is authenticated
 - the user has permission to view that page


vkurup (approved):
:shipit:

PR #109, opened 2021-03-26
HP-19 Add Posters

vkurup:
https://caktus.atlassian.net/browse/HP-19

QA instructions:
* Go to the Snippet Admin and add 1 or more Poster Categories
* Create a Poster List Page as a child of the Home Page and place it at the URL /posters/
* Create one or more Poster Detail Pages and they should show up in the proper category on the Poster List page
* Poster thumbnail is required. Clicking on it downloads the "main_poster"
* Each Poster can have multiple "additional_versions" and they will be listed below the poster thumbnail
* Categories are ordered alphabetically and Posters alphabetically
* I added disease-specific posters to the disease&condition page (The mockup for this wasn't quite the same, so I took a little liberty)


dchukhin (commented):
Good work on it!
I have 1 minor comment, then 🚢

PR #112, opened 2021-03-29
[HP-88] Data Reports Detail Page

dchukhin:
https://caktus.atlassian.net/browse/HP-88

This pull request:
 - gives CMS users ability to add a 4-column table in the Wagtail admin
 
 Now, CMS users should be able to create report detail pages in the Wagtail admin.
 
 TODO: add an archive page, but this depends on feedback from the client/Erik about what the archive page should look like.


Audiosutras (approved):
👍

PR #113, opened 2021-03-30
[HP-88] Archived Reports

dchukhin:
https://caktus.atlassian.net/browse/HP-88

This pull request:
 - adds a `DataReportDetailArchiveListPage` model, which uses a template with a list of all of its children
 - adds a `DataReportDetailArchiveDetailPage` model, which uses a template based on the `StaticPage`
 - removes the back button from the `DataReportDetailPage` template
 - shows the left navbar in the `DataReportDetailPage` template


Audiosutras (approved):
👍

PR #115, opened 2021-03-31
[HP-135] Give Icon Same Height As Text

dchukhin:
https://caktus.atlassian.net/browse/HP-135

This pull request removes the 'is-size-7-touch' CSS class from the contact info icon on the Closed POD homepage, so its height matches the height of the button with the text.


Audiosutras (approved):
👍

PR #116, opened 2021-04-01
[HP-121] Part I: Create ListPage

dchukhin:
https://caktus.atlassian.net/browse/HP-121

This pull request adds a generic `ListPage` model to the project, which can be used throughout the project to create pages to look like a list of other pages.
This pull request does not change any existing pages to use the `ListPage`, though that work can be done later.

PR #117, opened 2021-04-01
[HP-121] Add More StaticPage Table Types

dchukhin:
**This pull request depends on #116**

https://caktus.atlassian.net/browse/HP-121

In order to match the wireframes, the `StaticPage` is now able to support tables with 2, 3, 4, or 5 columns.


Audiosutras (commented):
1) Should we prevent the `ListPage` from being a direct child of the `HomePage`?
2) Should only 1 `static page` with a `slug` of `report-a-disease` exists as a direct child of `Homepage`?

apps/hip/models.py
```diff
@@ -184,6 +274,75 @@ def get_context(self, request):
...
+    show_header_in_right_nav = blocks.BooleanBlock(
+        required=False,
+        default=True,
+        help_text="Should this header be shown in the navigation on the right side of the page?",
+    )
+    rows = ListRowStreamBlock()
+
+
+class ListPage(HipBasePage):
+    show_breadcrumb = models.BooleanField(
```
Audiosutras, line 307: I guess I'm still a bit foggy on where the `ListPage` is to be made available and when this page is a child of another list page or specifically Homepage for that matter, how are are these ad hoc pages to be represented within the parent page's template. For example if a ListPage is a child of the Homepage, then should the ListPage show up in the sidebar for navigation?

Audiosutras: I'm wondering if a larger team discussion is needed to answer this? Overall, I really like your implementation!

dchukhin: Cool, thanks!
I think the idea is that the `ListPage` can be available at any level of the hierarchy, and it's up to the CMS editors to make things look the way they want. The left navbar is not currently editable by CMS users, and will continue to not be editable by them. I'm happy to have a team discussion about it as well.
For example, CMS users may want to choose what shows up in the left navbar, and I think Wagtail already offers fields for that, but this project disables them. Maybe those fields are a better way to make it clear what shows up in the left navbar?
Realistically, though, we'll probably run out of time before making it work entirely intuitively

dchukhin:
Thanks for the review!
To answer the questions:
1.) I don't see a need to prevent a `ListPage` from being a direct child of the `HomePage` for now. It seems like there's an implicit assumption that the direct children of the `HomePage` show up in the left nav. This certainly makes sense, though I don't know if our client expects that or not. Letting `ListPage` be a direct child of the `HomePage` could make it easier for the client to create any future pages at that level of the hierarchy. But 🤷, we can have a team discussion and change it if people feel strongly about it.
2.) I think Django raises an error if any 2 `Page`s have the same slug, so I don't think we need to do anything special about that

Audiosutras:
Ok, that sounds good.

Audiosutras (approved):
👍

PR #118, opened 2021-04-01
[HP-132] Don't Limit Paragraph Width

dchukhin:
https://caktus.atlassian.net/browse/HP-132

This pull request takes away the limit for a paragraph's maximum width. After checking all affected pages (`DataReportDetailPage`, `PCWMSAHomePage`, `ClosedPODHomePage`, also a number of pages with a right-side navbar, like the `StaticPage`), it seems that paragraphs are already limited in their width by their parent, so this restriction doesn't seem necessary.


Audiosutras (approved):
👍

PR #119, opened 2021-04-01
[HP-139] Report Detail Page Rightnav Styles

dchukhin:
This pull request:
 - moves the 'before_staticpage_content' block to within the main column of the `StaticPage`
 - moves the CSS class for the report detail and report archive detail pages to be on the main column of the template, rather than a parent of the columns. This way, the styles are only applied to the main column, and not the right nav.


Audiosutras (approved):
👾

PR #122, opened 2021-04-02
HP-133

Audiosutras:
https://caktus.atlassian.net/browse/HP-133

### DIFF
--- a/hip/static/styles/pages/data_report_archive_detail_page.scss
+++ b/hip/static/styles/pages/data_report_archive_detail_page.scss
@@ -1,5 +1,5 @@
 @import "../mixins";
 
-.reports-archive-detail-page-hip {
+.reports-archive-detail-page-mainarea-hip {
   @include wagtail-richtext;
 }
--- a/hip/static/styles/pages/data_report_archive_detail_page.scss
+++ b/hip/static/styles/pages/data_report_archive_detail_page.scss
@@ -1,6 +1,5 @@
 @import "../mixins";
 
 .reports-archive-detail-page-hip {
-  @include headers;
-  @include paragraphs;
+  @include wagtail-richtext;
 }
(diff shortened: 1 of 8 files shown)


hip/static/styles/pages/data_report_detail_page.scss
```diff
@@ -1,6 +1,5 @@
 @import "../mixins";
 
-.reports-detail-page-mainarea-hip {
-  @include headers;
-  @include paragraphs;
+.reports-detail-page-hip {
```
dchukhin, line 3: It looks like `reports-detail-page-hip` is not used in any HTML template, so these styles aren't applied on the report detail page

hip/static/styles/pages/data_report_archive_detail_page.scss
```diff
@@ -1,6 +1,5 @@
 @import "../mixins";
 
-.reports-archive-detail-page-mainarea-hip {
-  @include headers;
-  @include paragraphs;
+.reports-archive-detail-page-hip {
```
dchukhin, line 3: It looks like `reports-archive-detail-page-hip` is not used in any HTML template, so these styles aren't applied on the report archive detail page

dchukhin (commented):
Looks good for most of the files affected. After fixing the report ones, it should be ready to ➡️

hip/static/styles/pages/data_report_detail_page.scss
Audiosutras: Looks like I overwrote this class when I was copying and pasting. Great catch.

Audiosutras: DONE

hip/static/styles/pages/data_report_archive_detail_page.scss
Audiosutras: FIXED

Audiosutras:
Now excluding breadcrumb from list mixin.

dchukhin (approved):
📃

PR #123, opened 2021-04-02
[HP-138] Format External Report Dates

dchukhin:
https://caktus.atlassian.net/browse/HP-138

This pull request changes the external reports' `last_updated` value sent to the template to be a date, to match internal reports' `last_updated` value.

### DIFF
--- a/apps/reports/models.py
+++ b/apps/reports/models.py
@@ -1,3 +1,5 @@
+from datetime import datetime
+
 from django.db import models
 
 from wagtail.admin.edit_handlers import FieldPanel, PageChooserPanel, StreamFieldPanel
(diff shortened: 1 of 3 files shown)


apps/reports/models.py
```diff
@@ -73,8 +75,11 @@ def get_context(self, request):
         ]
         external_reports = []
         for report in self.external_reports.raw_data:
-            report_data = report["value"]
+            report_data = report["value"].copy()
             report_data["external"] = True
+            report_data["last_updated"] = datetime.strptime(
+                report["value"]["last_updated"], "%Y-%m-%d"
```
Audiosutras, line 81:
```suggestion
                report["value"]["last_updated"], "%b-%m-%d"
```
I think we want the abbreviated month instead of the full month. https://www.figma.com/proto/OZAfVKxh8fs8XE7RLnlbY7/PDPH-COVID19?node-id=1003%3A701&viewport=1007%2C-8587%2C0.4866814613342285&scaling=min-zoom. If so, I think we need to update internal links as well.

dchukhin: This line just takes a string like "2020-01-01" and turns it into a `datetime.date` object; then, the date is sent to the template. The formatting itself is done in the template, so I'll change it there.

Audiosutras: I see. thanks

Audiosutras (approved):
🍎

PR #124, opened 2021-04-02
HP-137 Decrease horizontal padding for poster titles

Audiosutras:
https://caktus.atlassian.net/browse/HP-137

### DIFF
--- a/hip/static/styles/pages/poster_list_page.scss
+++ b/hip/static/styles/pages/poster_list_page.scss
@@ -3,5 +3,5 @@
   @include lists;
   @include a-tags;
 
-  padding: 0.75rem 0.25rem;
+  padding: 0.75rem 2rem;
 }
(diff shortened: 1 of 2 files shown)


hip/static/styles/pages/poster_list_page.scss
```diff
@@ -3,5 +3,5 @@
   @include lists;
   @include a-tags;
 
-  padding: 0.75rem 2rem;
+  padding: 0.75rem 0.25rem;
```
dchukhin, line 6: It looks like this change only affects the padding on the main column, rather than on each poster card. It looks to me like it's the `card-content` padding that's causing the issue on this page

Audiosutras: Got it, will fix.

Audiosutras: done

dchukhin (approved):
🖼️

PR #128, opened 2021-04-05
Get Sidebar Authenticated Page Links From Context

dchukhin:
This pull request changes authenticated page URLs in the sidebar to come from the `authenticated_home_pages` context variable, rather than hard-coding it.


vkurup (approved):
:confetti_ball:

PR #130, opened 2021-04-05
[HP-106] Add Big Cities Home Page

dchukhin:
https://caktus.atlassian.net/browse/HP-106

This pull request:
* adds a `BigCitiesHomePage` page
* adds a method to HipBasePage for determining if a page is a Big Cities page or not
* adds a get_bigcities_home_page_url() utility function to return the `BigCitiesHomePage` URL
* redirects some users to the Big Cities homepage when they log in, according to the following order of precedence: if they are in the 'Closed POD' `Group`, then they go to the `ClosedPODHomePage`. Else, if they are in the PCW MSA `Group`, then they go to the `PCWMSAHomePage`. Else, if they are in the Big Cities `Group`, then they go to the `BigCitiesHomePage`.
* updates the left sidebar to show different options when the user is on a Big Cities page
* updates the left sidebar on the public site to show a link to the Big Cites home page on for users who have access to Big Cites part of the site

Note for testing: users are only redirected to the Big Cities home page upon login if they are in the ‘Big’ Cities’ Django `Group`. To test this functionality, add yourself to the 'Big’ Cities' `Group` (through the wagtail interface or the command line).


apps/auth_content/templates/auth_content/big_cities_home_page.html
```diff
@@ -0,0 +1,15 @@
+{% extends 'hip/static_page.html' %}
+
+{% block sidebar %}
+  {% include 'includes/sidebar.html' %}
+{% endblock %}
+
+{% block page_css_class_class %}bigcities-homepage-hip{% endblock %}
```
vkurup, line 7: Guess I didn't notice this before, but why does this block have 2 `class`es in it?

vkurup (approved):
:cityscape:

apps/auth_content/templates/auth_content/big_cities_home_page.html
dchukhin: Good question. I'd like to know as well. I'll change it to have just 1 `class`

PR #131, opened 2021-04-05
Test Fixtures File

dchukhin:
This pull request create a `conftest.py` file of fixtures from the `common` app, so they can be used throughout the test suite.


vkurup (approved):
Cool, it's things like this that I think can make pytest more powerful. Once one dev (i.e. Dmitriy) creates a useful fixture, this makes it much easier for other devs to reuse.

PR #132, opened 2021-04-05
HP-145: Don't try to validate file if it hasn't changed and turn off error emails

vkurup:
https://caktus.atlassian.net/browse/HP-145

First commit turns of error emails since we get them through sentry (thought I did this before, but obviously I didn't!)

Second commit changes our Document form to only validate the file if it has changed. If a user is only editing the title, for example, we shouldn't even try to validate the existing file.


dchukhin (approved):
📃

PR #133, opened 2021-04-06
[HP-141] Fixed Layout For Tables

dchukhin:
https://caktus.atlassian.net/browse/HP-141

This pull request:
 - adds a fixed `table-layout` to the `tables` mixin
 - uses the `tables` mixin on pages with tables throughout the site

### DIFF
--- a/hip/static/styles/_mixins.scss
+++ b/hip/static/styles/_mixins.scss
@@ -69,6 +69,7 @@
 @mixin tables {
   table {
     width: 100%;
+    table-layout: fixed;
   }
 }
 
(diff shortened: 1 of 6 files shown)


vkurup (approved):
This is definitely an improvement site wide!

My slight concern is that some of the tables (like the health alerts table) now have equal-width columns but the mockup shows that it should have 1 wide column with 2 narrower column. But IMHO, it still looks fine. So, I say that we merge this and ask Kat to look at all tables and see if we need to make tweaks to any tables. This is a much better default.

dchukhin:
Makes sense; thanks for the review!

PR #129, opened 2021-04-05
Allow PCWMSAHomePage To Have Children

dchukhin:
This pull request explicitly defines `PCWMSAHomePage`' children to be either `StaticPage`s or `ListPage`s.

PR #134, opened 2021-04-06
Allow ListPage To Have Children

dchukhin:
Similar to #129, this pull request explicitly defines `ListPage`s' children to be either `StaticPage`s or `ListPage`s.


vkurup (approved):
:children_crossing: :children_crossing:

PR #136, opened 2021-04-06
Update DiseaseControlListPage Children

dchukhin:
`DiseaseControlListPage`s may now have children that are `DiseaseAndConditionListPage`s, `StaticPage`s, or `ListPage`s. Also, there may only be 1 `DiseaseAndConditionListPage` for a `DiseaseControlListPage`.


vkurup (approved):
:petri_dish:

PR #139, opened 2021-04-07
HP-150 Highlight Sidebar Links For Nested Pages

dchukhin:
https://caktus.atlassian.net/browse/HP-150

This pull request updates the `getParentSidebarLink()` function to only return a sidebar element if it is found.


vkurup (approved):
:barber:

PR #142, opened 2021-04-07
[HP-151] Consistent Spacing Between Left Sidebar And Main Content

dchukhin:
https://caktus.atlassian.net/browse/HP-151

This pull request tries to make the space between the left sidebar and the main content of a page consistent across the site, so the breadcrumb element can be consistently spaced, and so that the pages look consistent to each other.


vkurup (approved):
Just had a question about the removal of the breakpoint-conditionals. Do we no longer need them because the new chosen padding is small enough that it works on both desktop and mobile? If so, :rocket: !

dchukhin:
Thanks for the review! I think that's correct about the breakpoint conditionals. The padding is smaller, and I couldn't get things to look bad with it, though we can see what happens in QA

PR #143, opened 2021-04-07
Limit Table Width

dchukhin:
This pull request limits table width to 80vw, so tables can not be bigger than the screen.

### DIFF
--- a/hip/static/styles/_mixins.scss
+++ b/hip/static/styles/_mixins.scss
@@ -70,6 +70,7 @@
   table {
     width: 100%;
     table-layout: fixed;
+    max-width: 80vw;
   }
 }
 


hip/static/styles/_mixins.scss
```diff
@@ -72,6 +72,7 @@
     table-layout: fixed;
     border-spacing: 2px;
     border-collapse: separate;
+    max-width: 80vw;
```
vkurup, line 75: This shrinks the width of the Health Alerts table on mobile a little bit. Could we bump it to 90vw? That value doesn't limit the health alerts table, but not sure if it achieves your goal on the pages you were looking at?

dchukhin: Thanks for the review. I was trying to fix the issue for tables that have overflowing content in the cells (like really long words in all of the columns), but I think this is actually fixed in #133, due to how `table-layout: fixed` determines cell width. I'm going to close this for now.

PR #146, opened 2021-04-09
Update dependencies to satisfy dependabot

vkurup:
https://github.com/caktus/philly-hip/security/dependabot

This updates the Python packages that dependabot is complaining about.


dchukhin (approved):
🆗

PR #148, opened 2021-04-09
HP-146

Audiosutras:
This PR removes `height` from `no-results-hip` class.  This will prevent the search modal from being very long when there are no search results.

https://caktus.atlassian.net/browse/HP-146


vkurup (approved):
:scroll:

PR #152, opened 2021-04-12
[HP-58] Part I: Backend Model & Views

dchukhin:
https://caktus.atlassian.net/browse/HP-58

This pull request:
 - adds a `ClosedPODContactInformation` model in the backend
 - displays instances of `ClosedPODContactInformation` in Wagtail under "Settings" by using `wagtail.contrib.modeladmin.options.ModelAdmin` for the `ClosedPODContactInformation` model
 - adds a `closedpod_contact_information` view to show a user's `ClosedPODContactInformation` data
 - adds a `closedpod_contact_information_edit` view to edit a user's `ClosedPODContactInformation` data


vkurup (approved):
:closed_umbrella:

Audiosutras:
This may be done as part of the work I'm doing, just wanted to mention this here. We need to pass the authenticated context as a part of the contact information views so that the sidebar does not show the unauthenticated links.

dchukhin:
Good point. I think that should be available in the template context as `authenticated_home_pages`, though the sidebar.html may need to be updated. Sure, let's do that as a part of the template.

PR #153, opened 2021-04-12
[HP-41] Allow user to add social media links to the second column of any tabl…

Audiosutras:
https://caktus.atlassian.net/browse/HP-41

This PR allows users to add social media links to the second column of any table. Table columns can vary from 2 to 5 columns.

Navigate to `Snippets` and create a `Social Media` instance. Then navigate to a `Static Page`, in this case `Related Websites` was created using a static page. Underneath a `table row`, the user is now able to use the Social Media snippet to add an optional social media instance to that row.


dchukhin (commented):
This works great for column 2, though I wonder if we could also let the CMS users choose which column to display to icons in. I imagine it won't always be in column 2. Maybe we can 🍐 on finding a solution for that?

Audiosutras:
Ready for your re-review.

dchukhin (approved):
🔗

PR #154, opened 2021-04-13
[HP-157] Replace HeatIndexPage And VolunteerPage With StaticPage

dchukhin:
https://caktus.atlassian.net/browse/HP-157

Wagtail users may now add children of the `EmergencyResponsePage` by using the `StaticPage`.


Audiosutras (approved):
👍

PR #159, opened 2021-04-15
Superusers Pass closedpod_user_check

dchukhin:
This pull request gives superusers permission to the Closed POD contact information page, regardless of whether they are in the "Closed POD" `Group` or not, by making them pass the `closedpod_user_check`.


vkurup (approved):
:superhero_woman:

PR #161, opened 2021-04-16
[HP-104] Partial Document Search

dchukhin:
https://caktus.atlassian.net/browse/HP-163

This pull request allows partial matching for searches on documents by:
 - creating a custom document model (`HIPDocument`)
 - using a custom manager for `HIPDocument` (`HIPDocumentQuerySet`) so that we can define our own `search()` method, which simply filters `HIPDocument`s case-insensitively based on their titles

Note: moving to a new document model means that all previous documents on the site get hidden, since we're no longer using that the same model. I used a fixture to load `wagtail.documents.models.Document` data into the `HIPDocument` database table.

[HP-104]: https://caktus.atlassian.net/browse/HP-104


apps/hip/models.py
```diff
@@ -541,3 +542,40 @@ def get_context(self, request):
...
+        matching.
+        """
+        # Split the search terms based on spaces.
+        search_terms = query.split(" ")
+        # Loop through the search_terms, and add to q_terms. All of the search
+        # terms have to be present for a HIPDocument to be matched.
+        q_terms = models.Q()
+        for term in search_terms:
+            if term:
+                q_terms &= models.Q(title__icontains=term)
```
vkurup, line 571: I don't think I've seen `&=` before. very cool!

vkurup (approved):
Looks good to me! How should we plan to get the documents moved to the new model on staging?

dchukhin:
Thanks for the review! I thought about writing a migration, but decided it's probably not worth it, and I was just going to use a fixture of the current documents.

PR #164, opened 2021-04-19
[HP-160] Update is_pcwmsa_page method

Audiosutras:
https://caktus.atlassian.net/browse/HP-160

### DIFF
--- a/apps/common/models.py
+++ b/apps/common/models.py
@@ -79,8 +79,17 @@ class HipBasePage(Page):
         A page is a PCW MSA page if it is either the PCWMSAHomePage, or a
         descendant of the PCWMSAHomePage.
         """
-        if hasattr(self, "pcwmsahomepage") or any(
-            [hasattr(ancestor, "pcwmsahomepage") for ancestor in self.get_ancestors()]
+        # Note: the PCWMSAHomePage inherits from StaticPage, which inherits from
+        # HipBasePage, so we must check self.staticpage.pcwmsahomepage and
+        # ancestor.staticpage.pcwmsahomepage.
+        if (
+            hasattr(self, "staticpage") and hasattr(self.staticpage, "pcwmsahomepage")
+        ) or any(
+            [
+                hasattr(ancestor, "staticpage")
+                and hasattr(ancestor.staticpage, "pcwmsahomepage")
+                for ancestor in self.get_ancestors()
+            ]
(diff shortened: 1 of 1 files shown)


apps/common/models.py
```diff
@@ -79,8 +79,17 @@ def is_pcwmsa_page(self):
         A page is a PCW MSA page if it is either the PCWMSAHomePage, or a
         descendant of the PCWMSAHomePage.
         """
-        if hasattr(self, "pcwmsahomepage") or any(
-            [hasattr(ancestor, "pcwmsahomepage") for ancestor in self.get_ancestors()]
+        # Note: the PCWMSAHomePage inherits from StaticPage, which inherits from
+        # HipBasePage, so we must check self.staticpage.pcwmsahomepage and
+        # ancestor.staticpage.pcwmsahomepage.
```
vkurup, line 84: You caught on to the main problem, but I think there's a slightly better solution. If you look at `self.get_ancestors()` it looks like `PCW MSA Secure Site` is in that list, but as a `Page` object, not a `PCWMSAHomePage` object, so we need to be checking `ancestor.specific` instead of `ancestor`.

So, something like:
```
modified   apps/common/models.py
@@ -80,7 +80,7 @@ class HipBasePage(Page):
         descendant of the PCWMSAHomePage.
         """
         if hasattr(self, "pcwmsahomepage") or any(
-            [hasattr(ancestor, "pcwmsahomepage") for ancestor in self.get_ancestors()]
+            [hasattr(ancestor.specific, "pcwmsahomepage") for ancestor in self.get_ancestors()]
         ):
             return True
         return False
```

Audiosutras: https://github.com/caktus/philly-hip/blob/aa546c20d3c4ba4009f3e1984f21635049fd63b8/apps/common/models.py#L88-L110
@dchukhin , does this method need to be updated as well?

dchukhin: Yes, I think so. It would be great to write a test also that proves that the current methods need fixing.

dchukhin: Good 👁️, Vinod

vkurup (approved):
Looks good, I guess we can make the other change you identified and the test updates that @dchukhin recommended in a future PR.

dchukhin:
Sounds good to me

PR #166, opened 2021-04-19
HP-144 Set up SSO

vkurup:
https://caktus.atlassian.net/browse/HP-144

This sets up SSO (in the most minimal way possible), using the client/tenant/secret provided by Graham. (The CLIENT_SECRET is also present in the LP entry "PDPH Philly HIP Secrets")

This will probably depend on them changing the redirect_uri as I mentioned in the call.

Improvements we should definitely talk about doing in future PRs or WRs:
* Figuring out how to change the Django/Wagtail admin pages
* Improve the current login template to be clearer for City and non-City users
* Update creds for production (once we have them)
* Other things?


hip/settings/base.py
```diff
@@ -152,6 +155,18 @@
     },
 ]
 
+# Azure auth using Python Social Auth
+# https://python-social-auth.readthedocs.io/en/latest/backends/azuread.html
+SOCIAL_AUTH_JSONFIELD_ENABLED = True
+SOCIAL_AUTH_AZUREAD_V2_TENANT_OAUTH2_KEY = os.environ.get("AZURE_CLIENT_ID")
```
dchukhin, line 161: How does this work with the `V2` in the setting name? I don't see that setting documented anywhere, though I do see `SOCIAL_AUTH_AZUREAD_TENANT_OAUTH2_KEY`

dchukhin (approved):
Looks 🆗 to me. Let's make improvements later

hip/settings/base.py
vkurup: Excellent :eye: ! I didn't realize that it wasn't documented. I found it because I was poring through the code trying to figure out why the email address wasn't properly populating on the new account and then I saw this backend: https://github.com/python-social-auth/social-core/blob/master/social_core/backends/azuread_tenant.py#L112 Specifically, this piece of code made sure that the email address was updated in our DB: https://github.com/python-social-auth/social-core/blob/master/social_core/backends/azuread_tenant.py#L121-L123 (I'll add a comment to the settings here)

dchukhin: Sounds good. Thanks!

PR #168, opened 2021-04-20
[HP-160] Additional Updates To is_pcwmsa_page() and is_bigcities_page()

dchukhin:
Following up on the comments in #164, this pull request:
 - updates `HipBasePage.is_bigcities_page` to be cleaner (like `HipBasePage.is_closedpod_page`)
 - updates tests for `HipBasePage.is_bigcities_page` and `HipBasePage.is_closedpod_page` to verify both children and grandchildren pages


apps/common/models.py
```diff
@@ -93,16 +96,12 @@ def is_bigcities_page(self):
         A page is a Big Cities page if it is either the BigCitiesHomePage, or a
         descendant of the BigCitiesHomePage.
         """
-        # Note: the BigCitiesHomePage inherits from StaticPage, which inherits from
-        # HipBasePage, so we must check self.staticpage.bigcitieshomepage and
-        # ancestor.staticpage.bigcitieshomepage.
         if (
             hasattr(self, "staticpage")
             and hasattr(self.staticpage, "bigcitieshomepage")
```
vkurup, line 101: I think this clause can be simplified to `if hasattr(self, "bigcitieshomepage")` now.

vkurup (approved):
One comment, then :ship: either way

apps/common/models.py
dchukhin: Makes sense; thanks

PR #171, opened 2021-04-20
HP-166 User can only edit their own info

vkurup:
https://caktus.atlassian.net/browse/HP-166

### DIFF
--- a/apps/auth_content/views.py
+++ b/apps/auth_content/views.py
@@ -72,7 +72,11 @@ def closedpod_contact_information_edit(request):
         # If the form is valid, then save it, and redirect the user to the
         # "closedpod_contact_information" view.
         if form.is_valid():
-            form.save()
+            instance = form.save(commit=False)
+            # 'user' field is not editable in the form since user is only able to edit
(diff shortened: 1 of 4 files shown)


apps/auth_content/views.py
```diff
@@ -72,7 +72,11 @@ def closedpod_contact_information_edit(request):
         # If the form is valid, then save it, and redirect the user to the
         # "closedpod_contact_information" view.
         if form.is_valid():
-            form.save()
+            instance = form.save(commit=False)
+            # 'user' field is not editable in the form since user is only able to edit
```
dchukhin, line 76: I wonder if we should raise an error in this case, like 'stop trying to edit other people's stuff!'

dchukhin (approved):
Looks good. I had 1 question, but I'm ok with having it work either way

apps/auth_content/views.py
vkurup: I thought about this as well, but I ended up thinking that since this can only be done maliciously, we shouldn't provide any information back to the user, but should instead just do what we would have done normally. The less information we give them, the better.

PR #172, opened 2021-04-20
HP-167 Remove some fields from "required" list

vkurup:
https://caktus.atlassian.net/browse/HP-167

I left Primary "provider" as required because cell-phone is required and "other" is a valid choice.


dchukhin (approved):
📴

PR #173, opened 2021-04-21
[HP-169] Footer Links

dchukhin:
https://caktus.atlassian.net/browse/HP-169

This pull request makes the links in the footer functional by:
 - setting footer links to URLs
 - setting the PDF URL to the value returned by a new context processor, `right_to_know_pdf_url`

Note: these links assume that the appropriate pages have been created in Wagtail.

From a user perspective, it would have been better to give Wagtail users the ability to set the footer URLs, but we're running out of hours, so they are hard-coded for now. Perhaps in the future these links can become editable in Wagtail.

### DIFF
--- a/apps/common/context_processors.py
+++ b/apps/common/context_processors.py
@@ -1,5 +1,7 @@
 from django.conf import settings
 
+from wagtail.documents import get_document_model
+
 from .utils import (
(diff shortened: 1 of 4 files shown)


vkurup (approved):
:foot:

PR #176, opened 2021-05-10
Fix GitHub Actions

dchukhin:
This pull request specifies 3.8.9 as our version of Python in our GitHub actions files, to avoid the regression in the 3.8.10 release (https://bugs.python.org/issue44061).


Audiosutras (approved):
💯

PR #177, opened 2021-05-13
DIS-1688 - Add google analytics 

Audiosutras:
https://caktus.atlassian.net/browse/DIS-1688

Please note (here)[https://standards.phila.gov/docs/development/analytics.html#analytics-tracking-for-webpages] bullet point 3.

This PR:
- updates `base.html` adding the google analytics script and iframe object. 
- Google Analytics should be present on each page.
- Analytics should work on `hip-staging.phila.gov`, but not `hip.caktus-built.com`.

### DIFF
--- a/hip/templates/base.html
+++ b/hip/templates/base.html
@@ -4,6 +4,16 @@
 <!DOCTYPE html>
 <html class="no-js" lang="en">
   <head>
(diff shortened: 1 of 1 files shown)


dchukhin (approved):
Seems like it should work

PR #178, opened 2021-05-13
DIS-1687 - Django-Import-Export 

Audiosutras:
https://caktus.atlassian.net/browse/DIS-1687

This PR:
- Adds `django-import-export` package
- Adds the `ExportMixin` to the `Users` and `HealthAlertsSubscribers` adminModel
- Django Admin users are now able to export users and healthalertsubscriber records in a variety of file formats.

### DIFF
--- a/apps/auth_content/admin.py
+++ b/apps/auth_content/admin.py
@@ -1,0 +1,31 @@
+from django.contrib import admin
+
+from import_export.admin import ExportMixin
+
+from .models import ClosedPODContactInformation
+
+
+@admin.register(ClosedPODContactInformation)
+class ClosedPODContactInformationAdmin(ExportMixin, admin.ModelAdmin):
+    list_display = (
+        "id",
+        "user",
+        "facility_name",
+        "facility_id",
+        "primary_contact_name",
+        "secondary_contact_name",
+    )
+    list_filter = (
+        "user",
+        "facility_name",
+        "facility_id",
+        "primary_contact_name",
+        "secondary_contact_name",
+    )
+    search_fields = (
+        "user",
+        "facility_name",
+        "facility_id",
+        "primary_contact_name",
+        "secondary_contact_name",
+    )
--- a/apps/auth_content/admin.py
+++ b/apps/auth_content/admin.py
@@ -14,7 +14,7 @@ class ClosedPODContactInformationAdmin(ExportMixin, admin.ModelAdmin):
         "facility_id",
     )
     search_fields = (
-        "user",
+        "user__email",
         "facility_name",
         "facility_id",
         "primary_contact_name",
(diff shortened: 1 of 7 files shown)


dchukhin (commented):
Cool, I like it!
A couple requests:
 - could we exclude the `password` field from the user export?
 - could we add this for the `ClosedPODContactInformation` model as well?

Audiosutras:
@dchukhin pushed comments to address your comments above. Thanks.

apps/auth_content/admin.py
```diff
@@ -0,0 +1,31 @@
...
+    list_display = (
+        "id",
+        "user",
+        "facility_name",
+        "facility_id",
+        "primary_contact_name",
+        "secondary_contact_name",
+    )
+    list_filter = (
+        "user",
```
dchukhin, line 19: I wonder if it would be a better user experience to not have any filters, since there will probably be a lot of submissions, each with its own `user`, `facility_name`, `facility_id`, `primary_contact_name`, and `secondary_contact_name`, which will:
- make this page take a long time to load
- make it difficult to find each of the filters

For example, if there are 100 users, I wouldn't want to scroll past all of them in order to find the facility name filter.

apps/auth_content/admin.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+@admin.register(ClosedPODContactInformation)
+class ClosedPODContactInformationAdmin(ExportMixin, admin.ModelAdmin):
+    list_display = (
+        "id",
+        "user",
+        "facility_name",
+        "facility_id",
+        "primary_contact_name",
```
dchukhin, line 15: Could we change the way the `primary_contact_name` and `secondary_contact_name` are displayed? Currently it's just "NAME*" and "NAME", which doesn't really explain what they are.
Or maybe it's easier to just not show them?

dchukhin (commented):
Looks good. I left a couple small comments about the `ClosedPODContactInformationAdmin`, then should be ready to 🛡️

apps/auth_content/admin.py
Audiosutras: Sounds good. Removed.

apps/auth_content/admin.py
Audiosutras: done

Audiosutras:
@dchukhin pushed changes

apps/auth_content/admin.py
```diff
@@ -0,0 +1,22 @@
...
+@admin.register(ClosedPODContactInformation)
+class ClosedPODContactInformationAdmin(ExportMixin, admin.ModelAdmin):
+    list_display = (
+        "id",
+        "user",
+        "facility_name",
+        "facility_id",
+    )
+    search_fields = (
+        "user",
```
dchukhin, line 17: It looks like this currently fails with "Related Field got invalid lookup: icontains". I think we probably want to search users' emails?
```suggestion
        "user__email",
```
Or maybe also `user__first_name` and `user__last_name`?

dchukhin (approved):
➡️

PR #179, opened 2021-05-14
[DIS-1670] Scan PDFs For Malicious Content

dchukhin:
https://caktus.atlassian.net/browse/DIS-1670

This pull request sets up PDF scanning when a user uploads a PDF file:
 - adds the `pdfid` library to the repo
 - adds a `scan_pdf_for_malicious_content()` utility function to determine if a PDF is suspicious, by scanning it with `pdfid`
 - calls `scan_pdf_for_malicious_content()` from the `ValidateFileTypeForm`
 - adds a custom `HIPDocumentAddView`, which forces the first file handler to be Django's `TemporaryFileUploadHandler`, so that we can give `pdfid` a file path to open the PDF file when it scans the PDF
 - sends requests to upload documents to `HIPDocumentAddView`, by setting it earlier than the Wagtail URLs in our `urls.py` configuration


Audiosutras (approved):
💯  Really nice job @dchukhin.

PR #180, opened 2021-05-17
[DIS-1689] Fix Data Reports Preview Error Bug

dchukhin:
https://caktus.atlassian.net/browse/DIS-1689

This pull request updates `DataReportListPage.get_context()` to handle the case that external reports' `last_updated` field is a date object, as well as the case that it is a string object.


Audiosutras (approved):
🥏

PR #181, opened 2021-05-18
[DIS-1678] Restrict SSO Users From Logging In With Username & Password

dchukhin:
https://caktus.atlassian.net/browse/DIS-1678

This pull request restricts Single Sign On users from logging in with a username & password by:
 - adding an `is_sso_user()` utility function that determines if a user is an SSO user
 - adding an `HIPAuthenticationForm`, which requires users to be non-SSO users when validating their data
 - using the `HIPAuthenticationForm` on the `HIPLoginView`
 - showing errors for each of the fields of the `HIPAuthenticationForm` in the front-end, rather than hard-coding an error message

### DIFF
--- a/apps/common/tests/test_utils.py
+++ b/apps/common/tests/test_utils.py
@@ -1,6 +1,9 @@
+import random
+
 from django.contrib.auth.models import AnonymousUser, Group
 
 import pytest
+from social_django.models import UserSocialAuth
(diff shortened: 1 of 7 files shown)


Audiosutras (approved):
💯 , please see DIS-1679 for the suggestion.

PR #183, opened 2021-05-19
[DIS-1679] Use Regular Login Page For Wagtail and Django Admin

dchukhin:
https://caktus.atlassian.net/browse/DIS-1679

This pull request redirects requests for the Wagtail login page and Django admin login page to go to the regular login page. The `next` URL parameter, if present, persists to the regular login page.

### DIFF
--- a/apps/hip/tests/test_views.py
+++ b/apps/hip/tests/test_views.py
@@ -255,3 +255,67 @@ def test_upload_document_post(db, client):
         # Verify that the HIPDocument was created.
         document = get_document_model().objects.get(title=expected_file_name)
         assert document.collection == Collection.get_first_root_node()
+
+
+@pytest.mark.parametrize("next_url", ["", "next_url"])
+def test_wagtail_login_redirects_to_cms_and_admin_login_get(db, client, next_url):
+    """Attempting to GET the Wagtail login redirects the user to the cms_and_admin_login view."""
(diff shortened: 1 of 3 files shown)


apps/hip/forms.py
```diff
@@ -31,3 +35,15 @@ def clean_file(self):
...
             return uploaded_file
+
+
+class HIPAuthenticationForm(AuthenticationForm):
+    def confirm_login_allowed(self, user):
+        """SSO users are not allowed to login using the HIPAuthenticationForm."""
+        # Run the AuthenticationForm's confirmation checks.
+        super().confirm_login_allowed(user)
+        # Raise an error if the user is an SSO user.
+        if is_sso_user(user):
```
Audiosutras, line 46: Suggestions rather than a requested change. Should we also see if the user has has 1) a phila email 2) is not an sso user yet. We could raise a validationError prompting the user to check with their IT department about getting an SSO login.

Audiosutras (approved):
Looks good. Added a suggestions for consideration, but I think this covers fully what the client is asking for.

apps/hip/forms.py
dchukhin: Good thinking. `is_sso_user()` returns `True` for any phila.gov email address, so this should already be handled, since the only way to get an account is to have someone else create one.

PR #184, opened 2021-05-19
[DIS-1680] Update Login Page Design

dchukhin:
https://caktus.atlassian.net/browse/DIS-1680

This pull request:
 - separates the login page vertically by using a divider `<div>` element, so that the username/password login is on the left side, and the SSO login is on the right side
 - updates the header link to say 'Log In', rather than 'Partner Log In'
 - updates the `partner_login_card.html` link to say 'Log In', rather than 'Partner Log In'
 - removes styles for the partner login card, which are no longer needed, since they were previously used to make sure the 'Partner Log In' button was not too wide

### DIFF
--- a/apps/hip/templates/includes/header.html
+++ b/apps/hip/templates/includes/header.html
@@ -64,7 +64,7 @@
               <span class="button is-size-7-touch header-icon-hip ml-0">
                 <i class="signin-icon-hip"></i>
               </span>
-              <span class="button is-size-7-touch header-btn-hip ml-0 pr-6">
+              <span class="button is-size-7-touch header-btn-hip ml-0">
                 <strong>Log In</strong>
               </span>
             </a>
--- a/apps/hip/templates/includes/header.html
+++ b/apps/hip/templates/includes/header.html
@@ -65,7 +65,7 @@
                 <i class="signin-icon-hip"></i>
               </span>
               <span class="button is-size-7-touch header-btn-hip ml-0 pr-6">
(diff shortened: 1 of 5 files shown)


hip/static/styles/common.scss
```diff
@@ -135,3 +135,16 @@ body.modal-is-open-hip {
...
   color: $alert;
 }
+
+
+.divider-vertical-hip {
+    display: flex;
+    flex-direction: column;
+}
+
+.divider-vertical-hip:before {
```
Audiosutras, line 145: Since we are modifying `.divider-vertical-hip` with the `before` psuedo-class I think we can nest this class inside of `divider-vertical-hip` using `&:before`.

apps/hip/templates/includes/header.html
```diff
@@ -65,7 +65,7 @@
                 <i class="signin-icon-hip"></i>
               </span>
               <span class="button is-size-7-touch header-btn-hip ml-0 pr-6">
```
Audiosutras, line 67: I think we should `center` the text for this button or make the button's width slightly smaller.

dchukhin: Sure, that sounds good. It looks like removing the extra padding on the right accomplishes this

hip/static/styles/common.scss
dchukhin: 🆗

Audiosutras (approved):
👍

PR #188, opened 2021-05-31
[DIS-1699] Update Partner Login Card Header Text

dchukhin:
https://caktus.atlassian.net/browse/DIS-1699

This pull request updates the header on the partner login card to be 'Log In', rather than 'Partner Log In'.


Audiosutras (approved):
💯

PR #190, opened 2021-06-07
[DIS-1674] Part III: Make Health Alerts Form More DRY

dchukhin:
https://caktus.atlassian.net/browse/DIS-1674

This pull request makes the health alerts page more DRY by having the view reuse the `generic_notification_signup` view. It also adds messages to our base template, so they can be displayed on any page.


Audiosutras (approved):
🧑‍💻

PR #195, opened 2021-06-15
[DIS-1701] Part III: Add 'next' URL Parameter For Forms

dchukhin:
https://caktus.atlassian.net/browse/DIS-1701

Currently, submitting the internal alerts form relies on the request's `HTTP_REFERER` parameter to determine the user's previous URL. This works ok, but in case the user submits the internal alerts form and gets the form back with errors, then the `HTTP_REFERER` header is set to the form's URL, meaning that submitting the form will bring the user back to the form.

This pull request: 
 - adds a `get_next_url_from_request()` utility function for getting the next URL from request parameters
 - uses the `get_next_url_from_request()` function to determine the success URL and close URL for the internal alerts form
 - adds a form `action` with the `next` URL parameter to the `notification_signup.html` template

### DIFF
--- a/apps/common/utils.py
+++ b/apps/common/utils.py
@@ -78,6 +78,20 @@ def get_all_pages_visible_to_request(request):
     return Page.objects.filter(id__in=pages_for_user.values_list("id", flat=True))
 
 
+def get_next_url_from_request(request):
+    """
+    Get the next URL from the request.
+
+    If the request has a 'next' parameter, then return its value. Otherwise,
+    return the value of the HTTP_REFERER header.
+    """
+    next_parameter = request.GET.get("next", "")
+    if next_parameter:
+        return next_parameter
+    else:
+        return request.META.get("HTTP_REFERER", None) or None
+
+
 def closedpod_user_check(user):
     """A check to determine if a user is in the ClosedPOD Group."""
     if user.is_superuser:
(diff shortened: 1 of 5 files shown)


apps/common/utils.py
```diff
@@ -78,6 +78,20 @@ def get_all_pages_visible_to_request(request):
     return Page.objects.filter(id__in=pages_for_user.values_list("id", flat=True))
 
 
+def get_next_url_from_request(request):
+    """
+    Get the next URL from the request.
+
+    If the request has a 'next' parameter, then return its value. Otherwise,
```
Audiosutras, line 85: Could we mention that `next` should be populated if being redirected from another url and that HTTP_Referrer is an optional header that gives us the url from where the request originated?

apps/notifications/tests/test_views.py
```diff
@@ -22,65 +23,65 @@ def test_get_internal_alerts_signup_page(db, client):
...
+    http_referrer_header,
+    expected_success_url,
 ):
     """
     POSTting valid data redirects the user, and shows a success message.
 
-    If the request has no HTTP_REFERER, then the get_emergency_communications_page_url()
-    utility function is used to determine where the user should be redirected upon
-    POSTing valid data.
+    If the request has a "next" parameter, then the user should be sent to that URL.
```
Audiosutras, line 47: Nice 👨‍🏫

Audiosutras (approved):
💯  awesome.

PR #200, opened 2021-06-22
[DIS-1708] Part I: Internal Alerts Form Requires More Fields

dchukhin:
https://caktus.atlassian.net/browse/DIS-1708

This pull request: 
 - adds 3 new fields to the `InternalAlertsSubscriber` model
 - makes the `InternalAlertsSubscriberForm` require one of the new fields when the `division` field has a certain value
 - hides each of the 3 new fields on the internal alerts subscriber page until the user chooses a value for the "Division" field which would require one of the new fields


Audiosutras (approved):
👨‍🏫

PR #182, opened 2021-05-18
Upgrade Some Python Dependencies

dchukhin:
This pull request upgrades:
 - some of our AWS-related dependencies to their latest bugfix versions
 - `django-debug-toolbar` to its latest bugfix version
 - `wagtail` to its latest bugfix version

PR #203, opened 2021-06-23
Upgrade Django To Version 3.2

dchukhin:
**Note: this pull request depends on #182, which should be reviewed first**

This pull request upgrades the Django to version 3.2.
 Based on the [release notes](https://docs.djangoproject.com/en/3.2/releases/3.2/), it does not look like anything in our code should break from this change. Being on version 3.2 means that this project will be on a Django version with long-term support until early 2024. Support for Django 3.1 ends at the end of 2021.


Audiosutras (approved):
👍

PR #204, opened 2021-06-24
[DIS-1715] Remove External Link Icon From URLs To Same Domain

dchukhin:
https://caktus.atlassian.net/browse/DIS-1715

This pull request updates the `addExternalLinkIconsAndDomain()` function in `common.js`, so that links to URLs on the same domain don't display the external link icon, even if they have the "external-linktype" CSS class.


Audiosutras (approved):
🔗

PR #206, opened 2021-06-24
Align Each Table Row In Wagtail At The Top

dchukhin:
In order to make the Wagtail interface a little easier to use, each row of tables are now aligned with each other at the top, so we avoid some table cells being above and others being below, like in the screenshot.
Screen Shot 2021-06-24 at 8 21 23 AM


Audiosutras (approved):
Quality of Life 🏆

PR #207, opened 2021-06-25
[DIS-1716] Keep Login Button Text Within Button

dchukhin:
https://caktus.atlassian.net/browse/DIS-1716

This pull request adds more specific rules about padding, so that the login button text fits within the button.


Audiosutras (approved):
🅰️

PR #209, opened 2021-07-23
[DIS-1669] Send Logs To AWS CloudWatch

dchukhin:
https://caktus.atlassian.net/browse/DIS-1669

This pull request:
 - updates the `apps.users.apps.UsersConfig` class to have a `ready()` method, which imports all of the signals in the `users` app. This is a change that we should have made when upgrading to Django3.2, but it looks like no one noticed that the signals stopped working, since we weren't sending the logs anywhere.
 - adds a task to install the "AWS for Fluent Bit" helm chart for the cluster, based on Colin's recommendation in our Slack channel. This task was run with `inv deploy.playbook deploy-cluster.yml`  
 
 The Django logs are now being sent to AWS CloudWatch, and can be seen here: https://console.aws.amazon.com/cloudwatch/home?region=us-east-1#logsV2:logs-insights .

### DIFF
--- a/apps/users/apps.py
+++ b/apps/users/apps.py
@@ -3,3 +3,7 @@ from django.apps import AppConfig
 
 class UsersConfig(AppConfig):
     name = "apps.users"
+
+    def ready(self):
+        # Import the signals here, so that they are registered.
+        from apps.users import signals  # noqa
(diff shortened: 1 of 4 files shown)


Audiosutras (approved):
👍

PR #214, opened 2021-08-20
Update Production SSO Client Secret

dchukhin:
This pull request updates the client secret for the SSO setup on the production server to the new value supplied by Aazim.
Note: this change has already been deployed to the production server.


Audiosutras (approved):
🍎

PR #215, opened 2021-08-20
Update Production SSO Secret Again

dchukhin:
The secret for SSO on the production server has been updated again, so I've encrypted the new value, and set it here.
Note: this change has already been deployed to the production server.


Audiosutras (approved):
🌡️

PR #217, opened 2021-09-15
WIP to add backups

Scottm-Caktus:
This is draft to add backups using the newest release of Kubase.


copelco:
Backups are created using postgres-client v12 and RDS is using v11:

https://github.com/caktus/k8s-caktus-backup/blob/36dd51d47d351d97588829ef6f1ab37a69ce61c5/Dockerfile#L24

The Dockerfile needs updates so it can perform a restore.

docs/hosting-services.md
```diff
@@ -0,0 +1,48 @@
+# Hosting Services
```
dchukhin, line 1: Nice, I like the documentation

Dockerfile
```diff
@@ -17,10 +17,13 @@ RUN set -ex \
     && RUN_DEPS=" \
     libpcre3 \
     mime-support \
-    postgresql-client \
+    postgresql-client-12 \
```
dchukhin, line 20: It looks like it's currently running version 11.12. Any concerns about updating to 12?
I guess either way now is a good time to do this, since production isn't live yet.

dchukhin (approved):
It works!
Thanks for setting these things up❗

PR #224, opened 2021-11-15
Fix Deployment Issue With RabbitMQ

dchukhin:
This pull request downgrades the version of `ansible-role-django-k8s` to 1.3.0, so deploys do not fail on a rabbitmq error:
```
failed: [staging] (item={'name': 'rabbitmq.yaml.j2', 'state': 'absent'}) => changed=false 
  ansible_loop_var: item
  item:
    name: rabbitmq.yaml.j2
    state: absent
  msg: Failed to find exact match for rabbitmq.com/v1beta1.RabbitmqCluster by [kind, name, singularName, shortNames]
```


Audiosutras (approved):
🍎

PR #232, opened 2022-01-04
Upgrade Django To Latest Bugfix Version

dchukhin:
To make sure that HIP isn't affected by the issues from today's Django [security release](https://www.djangoproject.com/weblog/2022/jan/04/security-releases/), this pull request upgrades Django to the latest bugfix version.
As a note, it looks like CVE-2021-45115 and CVE-2021-45116 are not actually issues for HIP, but it shouldn't hurt to be on the latest version.


s-putnam (approved):
Clicked through the site for a bit and everything looks good to me! 👍

PR #235, opened 2022-01-18
[DIS-1803] Shorter Documents URLs

dchukhin:
https://caktus.atlassian.net/browse/DIS-1803

This pull request makes it so that requesting a document means the user ends up on a page with a shorter URL by:
 - adding a new `get_document` view, which serves the file for a `HIPDocument` object
 - returning the URL for the `get_document` view in the  `HIPDocument.url` property
 
 When our HTML templates have a link that points to a document's `url` property, this link will now be served from our own `get_document` view.

Also, the following improvements were made:
 - the health alerts table only shows a link to a `HIPDocument` if the `HIPDocument` has a file; otherwise, just the text of the document title is shown
 - changes the `HealthAlertDetailPage.alert_file` relation to point to `hip.HIPDocument`. It looks like when we created the `HIPDocument` model, we forgot to update the `HealthAlertDetailPage.alert_file` field, so prior to this pull request, it was still pointing to `wagtaildocs.Document`. This pull request updates it to point to `hip.HIPDocument`.

One concern for serving Document files through Django would be a slow response, though it may not be an issue, since document files are typically only a couple pages long.


apps/hip/tests/test_views.py
```diff
@@ -319,3 +320,95 @@ def test_djangoadmin_login_redirects_to_cms_and_admin_login_post(db, client, nex
...
+def test_get_document_success(db, client):
+    """Sending valid parameters should return the expected document."""
+    document = DocumentFactory()
+    url = reverse(
+        "get_document",
+        kwargs={"document_id": document.id, "document_name": document.filename},
+    )
+    response = client.get(url)
+    assert response.status_code == 200
+    assert (
```
Audiosutras, line 334: This assertion failed for me locally.

```

    def test_get_document_success(db, client):
        """Sending valid parameters should return the expected document."""
        document = DocumentFactory()
        url = reverse(
            "get_document",
            kwargs={"document_id": document.id, "document_name": document.filename},
        )
        response = client.get(url)
        assert response.status_code == 200
>       assert (
            response.headers.get("Content-Disposition")
            == f'inline; filename="{document.filename}"'
        )
E       AttributeError: 'FileResponse' object has no attribute 'headers'
```

Audiosutras: It's possible that my db is just very outdated.

Audiosutras (commented):
I manually tested and everything seems to work well. I am stumped as to why the above test is failing locally for me. Could you could double check as well? I have reset my db and media files.

apps/hip/tests/test_views.py
dchukhin: Thanks for taking a look!
That's strange that it says a `FileResponse` wouldn't have a `headers` attribute, since [it's defined in Django's source code](https://github.com/django/django/blob/stable/3.2.x/django/http/response.py#L100) (`FileResponse` inherits from `StreamingHttpResponse`, which inherits from `HttpResponseBase`). Could you check the version of Django that you have installed locally? Maybe it's an outdated version where `FileResponse.headers` is not defined?

Audiosutras: My virtualenv for philly is on 3.1.8. updating resolved the issue.

Audiosutras (approved):
🌵

PR #236, opened 2022-03-03
DIS-1811 - AWS CloudWatch Alarm with SSO

s-putnam:
[Jira ticket](https://caktus.atlassian.net/jira/software/c/projects/DIS/boards/84?modal=detail&selectedIssue=DIS-1811)

Added code to base.py to handle SSO error


dchukhin (approved):
It looks like this change shows the Azure SSO error to the user in case Azure returns an error to our app.

PR #238, opened 2022-03-14
Hotfix: Update Production SSO Client Secret

dchukhin:
City of Philadelphia folks have been unable to log in to the production site, and it has been determined that the cause is an expired client secret. This pull request updates the client id and client secret to new values, which were provided by City of Philadelphia folks.

Note: to verify that the secret has been encrypted successfully, you should be able to run the following command:
`echo '<insert the secret here>' | ansible-vault decrypt && echo`

For example:
```
echo '$ANSIBLE_VAULT;1.1;AES256
32343734613830613937626264376338336539343135626262376634313664326330643762623836
3930353836316332656636336365643662383061623961390a303337326664303561633362386638
33633038383266393639663139366439346438333632343035336464346438656562383039363965
6163626634343364320a366663386465303166313835316137613838363631663130336362633034
31383130643730613139393861616231613665633461643936373866643530663861396266306332
6364656631353161623636336366356535653132363334353866' | ansible-vault decrypt && echo 
```


Audiosutras:
Looks good! 👍

Audiosutras (approved):
🍎

PR #239, opened 2022-03-16
Hotfix: Update Production SSO Client Id

dchukhin:
Following the work in #238, this pull request sets the `AZURE_CLIENT_ID` back to the value it had before #238, since it seems that the client id sent to us earlier this week is not actually being used; the old client id is being used.


s-putnam (approved):
Looks good! 📏

PR #240, opened 2022-04-01
Upgrade NodeJS from version 12 to version 16

s-putnam:
This PR upgrades NodeJS from version 12 to version 16. Changes include:
- updating and adding new dependencies and scripts in `package.json`
- removing `webpack_loader` dependency
- adding a new `gulpfile.js` file
- adding a new `webpack.config.js` file (and deleting 3 files- `webpack.local.config.js`, `webpack.prod.config.js`, and `webpack.base.config.js`)
- adding a `webpack_entry.js` file
- updating `Dockerfile` to use Node 16
- updating README to reflect NodeJS version change

To test locally:
- switch to using node 16
`nvm use 16`
- install new JS libraries
`npm install`
- run `make run-dev` and check
  - does editing JS files auto-generate a new bundle and show up locally?
  - does editing SCSS files auto-generate a new bundle and show up locally?
- run `npm run build` and check
  - does running the server (`python manage.py runserver`) have the correct JS and SCSS?
  - does editing JS files and running `npm run build` generate new bundle?
  - does editing SCSS files and running `npm run build` generate new bundle?
- click through site locally to determine it works like before (specifically in regards to JavaScript)


deploy/group_vars/all.yml
```diff
@@ -92,7 +92,7 @@ k8s_container_port: 8000
 k8s_container_image: "{{ RepositoryURL }}"
 k8s_container_image_pull_policy: Always
 k8s_container_replicas: 2
-k8s_migrations_enabled: true
+k8s_migrations_enabled: false
```
dchukhin, line 95: I think this was changed temporarily for a deploy; could we change it back to "true", so that migrations run on future deploys?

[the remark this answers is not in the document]
s-putnam: Sure, I got that changed and pushed!

gulpfile.js
```diff
@@ -0,0 +1,38 @@
...
+    [
+    './hip/static/styles/bundle.scss',
+    './hip/static/styles/**/*.scss'
+    ], gulp.series('styles')
+  );
+  done();
+});
+
+
+exports.default = gulp.series('styles', 'javascript', 'watch-css', 'watch-js');
```
Audiosutras, line 38: Can you add another line to the end of the file.

webpack.config.js
```diff
@@ -0,0 +1,31 @@
...
+            }
+        ]
+    },
+    plugins: [
+        new webpack.ProvidePlugin({
+            $: 'jquery',
+            jQuery: 'jquery',
+        }),
+    ]
+};
```
Audiosutras, line 31: Can we add a new line here.

gulpfile.js
```diff
@@ -0,0 +1,38 @@
+const gulp = require("gulp");
```
Audiosutras, line 1: This part is a bit confusing to me and I don't know; but why are we removing `webpack_loader` and switching to using gulp instead?

Audiosutras:
Fantastic work Sage. I have left a couple of comments.

gulpfile.js
dchukhin: I can say a little about that, though I'm happy to discuss the options before moving ahead.
This came about as a result of trying to upgrade webpack from v4 to v5, which caused things to break. In a separate conversation with Jeremy, I mentioned that `webpack_loader` seemed to not work with webpack v5 when starting a new project with jade-truffle, and he mentioned that SC had decided to stop using `webpack_loader` because they didn't need it, and have [an open pull request](https://github.com/caktus/jade-truffle/pull/32) to update jade-truffle to remove it, besides other changes.
Due to these things, I encouraged replacing `webpack_loader` in this project with the gulp approach from the jade-truffle pull request.

Let us know if you have concerns, or if keeping `webpack_loader` is better.

Audiosutras: Thanks for clarifying @dchukhin and that's good to know.

Audiosutras (approved):
👍 🥳

gulpfile.js
s-putnam: Done! 👍

webpack.config.js
s-putnam: And done! 🥳

PR #246, opened 2022-08-03
Ignore webpack-stats.json

s-putnam:
This PR adds `webpack-stats.json` to the `.gitignore` file, so that we don't accidentally commit it.


dchukhin (approved):
🙈

PR #252, opened 2022-09-28
[DIS-1801] Improves color contrast of right-hand nav and adds arrow indicator

s-putnam:
[Ticket](https://caktus.atlassian.net/browse/DIS-1801?atlOrigin=eyJpIjoiMjJkODI4NzdmMDhmNGJlMzhkNDhkODNhNzM3M2E5NjgiLCJwIjoiaiJ9)

This PR:
- Changes the font color for headings in the Right-hand nav (as on the Posters page) to provide better contrast with the white background
- Adds an arrow to indicate the active heading, as contrast between active heading color and non-active heading color was too low

Note: The arrow indicator was not requested, and can be removed or altered if requested. It or some other non-color-based indicator will be required for full accessibility

### DIFF
--- a/hip/static/styles/pages/static_page.scss
+++ b/hip/static/styles/pages/static_page.scss
@@ -39,9 +39,18 @@
 
 .right-scroll-link-hip {
   // links are grey unless they are current
-  color: $grey;
+  color: $dark-grey;
 
   &.is-current-hip {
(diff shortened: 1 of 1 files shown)


dchukhin (approved):
🟦

PR #259, opened 2022-12-20
update uwsgi and libffi-dev to build_deps

ronardcaktus:
Fix `cffi` and `uwsgi` from breaking during deployment.


jcjudkins (approved):
Approved!

PR #270, opened 2023-06-30
Avoid Server Error On Health Alerts Without Alert File

dchukhin:
ClickUp ticket: https://app.clickup.com/t/8684y6f05

Previously, the `HealthAlertDetailPage.alert_file` field was optional, but serving a health alert page redirected users to the `HealthAlertDetailPage`’s `alert_file`’s URL. As a result, loading a `HealthAlertDetailPage` with an empty `alert_file caused a server error.

Now, if a `HealthAlertDetailPage` has an empty `alert_file`, a generic 404 page will be returned (whether in the Wagtail preview panel, or on the site).

Note: it appears that this server error can only be triggered for Wagtail users. Regular (non-authenticated) users can click on the health alerts from the health alerts list page ([which only have a link if the `HealthAlertDetailPage` has an `alert_file`](https://github.com/caktus/philly-hip/blob/develop/apps/hip/templates/includes/health_alert_table_row.html#L5-L10).

Note: I first tried to make the `HealthAlertDetailPage.alert_file` field required, but it appears that a number of `HealthAlertDetailPage`s have an empty `alert_file` field, and that may be meaningful data (a health alert was issued, but we don’t have a copy of the alert), so I decided to leave the `alert_file` field optional, rather than trying to programmatically decide what should happen for those already-existing objects in the database.

### DIFF
--- a/apps/health_alerts/models.py
+++ b/apps/health_alerts/models.py
@@ -1,6 +1,7 @@
 import datetime
 
 from django.db import models
+from django.http import Http404
 from django.shortcuts import redirect
 
 from phonenumber_field.modelfields import PhoneNumberField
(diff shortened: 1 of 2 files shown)


ronardcaktus (approved):
Looks 🐅

PR #274, opened 2023-09-11
Update To Latest Django Bugfix Version

dchukhin:
This pull request upgrades Django to the latest bugfix version for Django3.2.


ronardcaktus (approved):
Looks 🦊

Mr. 🥋I thought it would be a good idea to only notify users of LTS releases on the Wagtail admin. If you disagree, feel free to remove it.

dchukhin:
That sounds great; it looks like there's a new bugfix release of Django 3.2, so we should update to it too

ronardcaktus:
> That sounds great; it looks like there's a new bugfix release of Django 3.2, so we should update to it too

I updated it to the latest.

dchukhin:
Thanks for making the change @ronardcaktus 🍄

PR #275, opened 2023-10-30
Avoid Memory Leaks [No ClickUp ticket]

ronardcaktus:
This PR enables reloading workers after a specified amount of managed requests - `1000` - to avoid memory leaks.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -183,3 +183,6 @@ ENV PATH=/code/venv/bin:$PATH
 WORKDIR /code
 
 CMD ["python", "/code/manage.py", "runserver", "0.0.0.0:8000"]
+
+# Reload workers after the specified amount of managed requests (avoid memory leaks)
+ENV UWSGI_MAX_REQUESTS=1000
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -183,3 +183,6 @@ ENV PATH=/code/venv/bin:$PATH
 WORKDIR /code
 
 CMD ["python", "/code/manage.py", "runserver", "0.0.0.0:8000"]
+
+# Reload workers after the specified amount of managed requests (avoid memory leaks)
+ENV UWSGI_MAX_REQUESTS=1000
```
dchukhin, line 188: Should we set this earlier on in the file, before we start uWSGI?

copelco: Ah yes, I think it should be up by line 89

ronardcaktus: I moved it up. Great suggestion @dchukhin. I moved it 🆙.

dchukhin (approved):
1️⃣0️⃣0️⃣0️⃣

PR #284, opened 2024-08-26
Run tests in GH runner and pin django-compressor

ronardcaktus:
After merging number #282 the deployment stopped working. This is because only certain IPs have access to the application (per the client's request), see which IPs have access [here](https://github.com/caktus/philly-hip/blob/f3f9832a3841a06f48b664555b524806c8dc8855/deploy/group_vars/all.yml#L20-L25). To fix tests and deployments, we created a Github runner in that same PR #282. 

This PR enables tests to run in that runner, thus enabling tests & deploys to work again.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -7,7 +7,9 @@ on:
 
 jobs:
   test:
-    runs-on: ubuntu-20.04
+    runs-on:
+      - self-hosted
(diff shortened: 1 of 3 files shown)


ronardcaktus:
Work continued here - https://github.com/caktus/philly-hip/pull/285#issuecomment-2386467706

PR #295, opened 2025-03-19
Update Django v5.2.1, Wagtail v7.0 (LTS) and minor fixes

ronardcaktus:
This PR does the following:

- Updates Django v4.2 -> ~~5.1~~ 5.2.1
- Updates Wagtail v5.2 -> ~~6.3~~ 7.0 (LTS)
- Replaced the `email` field for users from the deprecated Postgres field `CIEmailField` to a vanilla `EmailField`. See [here](https://adamj.eu/tech/2023/02/23/migrate-django-postgresql-ci-fields-case-insensitive-collation/) for more details on the deprecated `CIEmailField` and the update process.

**Notes:** 

Several things changed in both Django and Wagtail, which required changes in the codebase, most notably:
- Django: Django now uses a GET request for admin logout, therefore a log message is no longer produced. That required the modification of [this test](https://github.com/caktus/philly-hip/blob/94bf71f1e2dd3f99a2c76b3abb3d943dda62c74d/apps/users/tests/test_signals.py#L49-L69). See [here](https://forum.djangoproject.com/t/deprecation-of-get-method-for-logoutview/25533) and [here](https://code.djangoproject.com/ticket/15619) for more details.
- Wagtail: `partial_match` was deprecated in Wagtail. So I removed `partial_match=True`.

`wagtail_modeladmin` was separated into a stand-alone package.


copelco:
Task linked: [CU-868d44e5z Upgrade Django, Wagtail, and minor packages](https://app.clickup.com/t/868d44e5z)

ronardcaktus:
CI is failing, maybe I ran migrations before updating to Django 5.1 and Wagtail 6.3?
https://github.com/wagtail/wagtail/issues/11928#issuecomment-2096210151

philtuskha (commented):
Hi Ronard I searched around on https://adamj.eu/tech/2023/02/23/migrate-django-postgresql-ci-fields-case-insensitive-collation/ to attempt to solve the CI issues to no avail.  

I have an observation about the copyright.  Let me know if that looks like sound logic or not.

apps/hip/templates/includes/footer.html
```diff
@@ -1,6 +1,6 @@
 <footer class="footer-section-hip has-text-centered pt-5">
   <div class="footer-section-blue-hip">
-    <p class="pt-5">Copyright 2025 by The City of Philadelphia. All Rights Reserved.</p>
+    <p class="pt-5">Copyright 2015 - 2025 by The City of Philadelphia. All Rights Reserved.</p>
```
philtuskha, line 3: Looking at the wayback machine, it looks like 2015 was when the site was officially spun up and so there could be content that would need to be protected.

apps/users/migrations/0004_alter_user_email_field.py
```diff
@@ -1,24 +1,40 @@
...
             model_name="user",
             name="email",
             field=models.EmailField(
                 db_collation="case_insensitive", max_length=255, unique=True
             ),
         ),
+        migrations.RunSQL(
+            sql=f'CREATE INDEX "users_user_email_idx" ON "users_user" ("email")',
+            reverse_sql='DROP INDEX IF EXISTS "users_user_email_idx"',
+        ),
```
philtuskha, line 39: It seems that the database is unhappy when you use `AlterField` when changing from CIEmail to Vanilla even with the collation.  And that this error was thrown `django.db.utils.NotSupportedError: nondeterministic collations are not supported for operator class "varchar_pattern_ops"` ...Reading a bit about this in this thread here:

https://code.djangoproject.com/ticket/33901#comment:18

An alternative solution would have been to drop the email field, do a data migration and then add a new field, but that seemed more dangerous.

PR #300, opened 2025-06-13
Navigation updates

philtuskha:
**What changed?**

This PR adds new `PageLink` and `Menu` snippets to the repo, which allows a wagtail user to edit the main menu and resources menu.  A migration was added to get the data of these two menus into the state that we see on production.  Note: other menus found in the sidebar when viewing other page types (`closedpod page`, `bigcities page` & `pcwmsa page`) could be changed to use this logic in the future, but are still static in this PR.    

**Why?**
The purpose of this PR is to allow either us or philly-hip to be able to change these menus more easily for this particular change and also in the future.  It also can prevent the possibility of broken links.

**QA?**

Run migrations locally with database as seen on staging or production.  Go to `/cms/snippets/` > `Menus`.  See new `Main menu` and `Resources Menu`.  Click in to edit titles and pages where they are linked.  The order of these menus can also be changed.

From email:

Community Resources (this would link [here](https://hip.phila.gov/emergency-response/php-community-resources/))
Newsletters
Posters
Related Websites

(Note: The conferences page would be moved elsewhere, or the individual events would be reassigned to their respective program pages.)

### DIFF
--- a/apps/hip/migrations/0026_menu_pagelink.py
+++ b/apps/hip/migrations/0026_menu_pagelink.py
@@ -138,6 +138,7 @@ class Migration(migrations.Migration):
                         on_delete=django.db.models.deletion.CASCADE,
                         related_name="+",
                         to="wagtailcore.page",
+                        help_text="Title is changed to page title when using this page chooser, but can be altered afterwards.",
                     ),
                 ),
                 (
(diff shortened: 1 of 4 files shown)


ronardcaktus (approved):
🏟️ ⚾️

PR #304, opened 2025-07-10
Fix 500 error when searching user list in Django admin

ronardcaktus:
This PR fixes the 500 error when searching for users in the Django admin. The 500 errors started happening because of the DB collation from CIEmailField to the vanilla EmailField. 
The search is not the best because it looks like while collated fields support `__iexact`, it does not support `__icontains` (partial search). Search works when it's an **exact** match.

Closes:
 - https://app.clickup.com/t/868ep6xyu

### DIFF
--- a/apps/users/admin.py
+++ b/apps/users/admin.py
@@ -31,6 +31,26 @@ class CustomUserAdmin(ExportMixin, UserAdmin):
         "user_permissions",
     )
 
+    def get_search_results(self, request, queryset, search_term):
+        if not search_term:
+            return queryset, False
+
+        email_results = queryset.filter(email__iexact=search_term)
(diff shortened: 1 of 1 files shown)


copelco:
Task linked: [CU-868ep6xyu 500 error when searching user list in Django admin](https://app.clickup.com/t/868ep6xyu)

ronardcaktus:
> I might suggest doing this instead:
> 
> https://github.com/caktus/publish-mdm/blob/c28a253b83e361fde57447a3248e802b1f358037/apps/mdm/admin.py#L308-L314
> 
> And then referencing that annotation in the `search_fields` list

That's cleaner and works better.

keanyaphelps (approved):

PR #318, opened 2026-02-04
Replace sign up link

ronardcaktus:
This PR replaces the sign up link with a new link provided by Brandon. He made the replacement request today and said that they need this deployed by Friday.
I asked him what he wanted to do about the old form, but he wanted to keep it for now. 

For more details, see https://github.com/orgs/caktus/projects/22/views/1?pane=issue&itemId=154550099&issue=caktus%7Cphilly-hip%7C317

### DIFF
--- a/apps/disease_control/templates/disease_control/disease_and_condition_detail_page.html
+++ b/apps/disease_control/templates/disease_control/disease_and_condition_detail_page.html
@@ -2,120 +2,122 @@
 {% load wagtailcore_tags %}
 
 {% block content %}
-  <div class="columns is-desktop">
-    <div class="column">
-      {% include 'includes/breadcrumb.html' %}
+<div class="columns is-desktop">
+  <div class="column">
+    {% include 'includes/breadcrumb.html' %}
 
-      <div class="disease-page-hip px-4">
-        <div class="nav-heading-hip" id="section-{{ page.title|slugify }}"></div>
-        <h2>{{ page.title }}</h2>
+    <div class="disease-page-hip px-4">
+      <div class="nav-heading-hip" id="section-{{ page.title|slugify }}"></div>
+      <h2>{{ page.title }}</h2>
 
-        {% if page.at_a_glance %}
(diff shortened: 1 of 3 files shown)


apps/disease_control/templates/disease_control/disease_and_condition_detail_page.html
```diff
@@ -2,120 +2,122 @@
...
+      <div class="nav-heading-hip" id="section-{{ " Health Alerts"|slugify }}"></div>
+      <div class="level">
+        <div class="level-left is-align-items-baseline is-block-desktop-only">
+          <div class="level-item">
+            <h2>Health Alerts</h2>
+          </div>
+          <div class="level-item pl-2">
+            {% if show_more_health_alerts %}
+            <a href="{% pageurl page.health_alerts.first.get_parent %}?condition={{ page.slug }}">View all alerts for {{
+              page.title }}</a>
```
kmtracey: This wrapping is going to break the template rendering I think? Can this file be updated wiht just the relevant change and not the whitespace changes?

kmtracey: Oh I guess it may not break, the pageurl tag is all on one line. Still the break in the variable brackets is weird and I don't know offhand if that will work or not.

ronardcaktus: It's my VS Code extension 😞. Let me disable that.

Done

kmtracey (approved):
By "they want to keep the old form" for now do they realize the old url will still work and accept submissions if somehow it's been replicated somewhere else? One possible source: in searching the code base I see there is a migration that added a redirect from ""/HealthAlerts/SignUpHealthAlerts" to the old view they are replacing -- that looks to be still in place so could be one way the old view is still accessed. Who will be responsible for removing/changing that redirect (not sure if you can redirect offsite via redirect...maybe you can...)

